PR #6, opened 2015-02-03
add package build via cmake

saces:
basic support for building deb/rpm/etc packages via cmake


MathJud:
Linux package build support

PR #21, opened 2015-05-23
Create LICENSES.md

deavmi:


MathJud:
Thanks for pointing out that the Licenses were missing. I copied the files from qaul.net. 
I prefer text only files, as they can be used directly for the installers.

deavmi:
@MathJud Okay, that's great. Thanks :smile:.

PR #30, opened 2015-05-24
Update README_Android.md

deavmi:


MathJud:
please change the following things:

* correct the project name typo 
* We always use the project name in lower case. even at the beginning of a sentence! To stress the fact that it is a web URL. It is used in an iconized / logo kind of way.

deavmi:
@MathJud Shall do this.

PR #36, opened 2015-05-24
Update README.md

deavmi:


MathJud:
I don't think this makes the page more informative. I personally like the old 'less text' version better ...

deavmi:
@MathJud okay. :smile:

PR #43, opened 2015-05-25
Update README_Android.md

XenGi:


MathJud:
Thanks for the suggestions! 
Your hint for readability of the document due to long devices list led to a split of the Android documentation, with a new file about device compatibility.

PR #95, opened 2016-02-06
small fixes to the cli client

jaromil:
gather the configured PREFIX from the build stage
use stderr for all logs so scripts can pipe and parse


MathJud:
Fixes to make qaul-cli client compatible with devuan distribution.

PR #102, opened 2016-03-05
cmake provides hint to the user to proceed

anuvazhayil:


MathJud:
Thanks for the pull request!
Unfortunately, the hint is only correct for the Linux operating system.
The commands are different on the following platforms:
* Windows
* Android

There should be a platform switch that shows the right commands when building on other platforms.

saces:
Is the text really blue? Why?
A '--no-color' option is necessary.

Since 'echo -e' is not POSIX it should be somehow avoided. At least on FreeBSD the outout looks quite ugly, the recorded output from autometed build machines will too...

MathJud:
Thank you very much for your great work!!!

PR #136, opened 2016-07-07
Do we still need directory 'archive/'?

deavmi:
Do we still need directory 'archive/'?


MathJud:
Yes, we need this directory!
During the cmake build process external sources such as pjsip or olsrd are downloaded and stored in this directory.

deavmi:
OH okay. Sorry. :)

PR #144, opened 2016-07-07
Code highlighting added.

deavmi:


MathJud:
Where do you see the advantage of this change?

deavmi:
Looks more processional.

deavmi:
* professional

MathJud:
You're right, there is a formatting error!
But it's the other way round: Everything indented should be formatted as code and not as titles, even when it has a '#' in front of it.

deavmi:
Okay.

deavmi:
Should I fix something here or close this PR?

MathJud:
It would be cool, if you like to do that. It would be great, everything that is indented should be wrapped as a whole in the markdown tag for code.

deavmi:
@MathJud I wanted to know. Is the iOS version still being developed?

deavmi:
What changes should I make to this file? @MathJud

MathJud:
I fixed this with commit 71cf777ff2f2cbd2a39ca18e90224be5233a564d

PR #147, opened 2016-07-09
Neater UI

deavmi:


MathJud:
Hmm ..., there is a new [logger](https://github.com/WachterJud/qaul.net/blob/qaul-ng/src/libqaul/include/qaullib/logging.h) which will replace all fprintf calls.
It shall be configurable from the command line. This is not realized yet, but the logger functions are here:
https://github.com/WachterJud/qaul.net/blob/qaul-ng/src/libqaul/include/qaullib/logging.h

deavmi:
I think that this logger should be implemented and then via the command-line the user should be able to give a command line argument with the given level of verbosity of the logs.

PR #148, opened 2016-07-12
Neater command-line code

deavmi:


MathJud:
thanks!
But I think it should explain, that this script is only for debian based distros, and I would like to have it in a manner, that users/developers can add the commands for other distros.

deavmi:
Okay. :)

PR #157, opened 2016-08-21
Is this acceptable?

deavmi:
They only are referneced in the for loops so why not declare them there as the syntax suggests so.


deavmi:
Oh I see now. C99 allows it but not the newer C11?

PR #242, opened 2019-06-11
Building out the API

NoraCodes:
I'm using this PR as a place to build out the API and the testing infrastructure. Here, I've defined the basic structure of the user-to-user message API and a way to detect errors in those messages.

This uses the BLAKE2 algorithm, which is of similar strength to SHA-3 (Keccak) but much faster to compute.


libqaul/common/src/message.rs
```diff
@@ -0,0 +1,19 @@
...
+/// A `Message` to be sent from the `sender` to the `recipient`, associated
+/// with the given `service`.
+/// Differs from a RATMAN message; this is a user-to-user, fully encapsulated
+/// message which does not care (or know) how it is delivered, merely that it
+/// was sent by one user and is meant for another.
+pub struct Message {
+    sender: UserID,
+    recipient: UserID,
+    // TODO DESIGN: Include a fingerprint here?
+    // TODO DESIGN: It's probably a good idea to add a signature in addition to
```
spacekookie: I would say "yes" to both of those.

Although firstly I guess we should answer the question: is the fingerprint and "UserID" different? From how I see it we have three IDs at play here

- qaul.net ID
- fingerprint
- RATMAN routing ID

We could just make all three the same. The potential upside is that it requires less space. Not sure if there's any downsides other than _maybe_ more likely collisions?

Once we figured this out, I would suggest we include the signature of some always-changing payload like a timestamp or message counter

libqaul/common/Cargo.toml
```diff
@@ -5,4 +5,5 @@ version = "0.1.0"
 authors = ["Katharina Fey <kookie@spacekookie.de>"]
```
spacekookie: Feel free to add yourself as an author too

libqaul/common/src/lib.rs
```diff
@@ -1,9 +1,16 @@
 //! A small library with data types used across qaul.net
+extern crate blake2;
+mod message;
+mod payload;
+
+pub type HashBytes = [u8; 32];  // Because of the lack of type-level generics,
```
spacekookie: I'm wondering if we maybe wanna use the [generic array](https://docs.rs/generic-array/0.13.0/generic_array/) crate here? I've never worked with it directly but a bunch of RustCrypto crates use it for exactly this reason

libqaul/common/src/payload.rs
```diff
@@ -0,0 +1,89 @@
...
+    <A as AsMut<[T]>>::as_mut(&mut a).copy_from_slice(slice);
+    a
+}
+
+/// The actual content of a message, along with the mechanism to validate that
+/// no transmission errors occurred (message digest).
+#[derive(PartialEq, Eq, Debug)]
+pub struct Payload {
+    // TODO DESIGN: add a payload type field?
+    // Upcast to u64 so that there is no issue knowing what the length of a Payload is when sent to different platforms,
```
spacekookie: We could totally do an ASN.1 and have variable length lengths :joy: (please ignore me)

Using a 64bit unsigned for length sounds like a good idea. Payload type could probably just be another string, similar to the service. Not something _we_ as library developers really care about but something where service authors can do stuff with. Good idea :+1:

spacekookie (commented):
This looks really good :) I've added a few comments with my thoughts

libqaul/common/src/message.rs
NoraCodes: I definitely think we should collapse as much as possible, but the qaul.net ID and the RATMAN ID are conceptually different, right?

libqaul/common/src/lib.rs
NoraCodes: If you're ok with making that our standard; it's hard to mix `GenericArray` with regular arrays.

spacekookie:
Funnily enough, not as much as you'd think.

RATMAN doesn't actually have the concept of a "computer node" in its routing. It only cares about mapping ids to specific network addresses.
And those ids could very well just be the user ids i.e. user fingerprints.

This also means that if someone wanted to get a new id on the network, they could easily do this and pretty much nobody on the network would be able to figure out that they're the same person

libqaul/common/src/message.rs
spacekookie: Funnily enough, not as much as you'd think.  RATMAN doesn't actually have the concept of a "computer node" in its routing. It only cares about mapping ids to specific network addresses. And those ids could very well just be the user ids i.e. user fingerprints.

This also means that if someone wanted to get a new id on the network, they could easily do this and pretty much nobody on the network would be able to figure out that they're the same person

(not sure what the email reply did here first 😅)

libqaul/common/src/lib.rs
spacekookie: Yea, sounds good to me

libqaul/common/src/message.rs
NoraCodes: OK, let's make them all the same.

libqaul/common/src/payload.rs
NoraCodes: After looking into this I think maybe the better place for this functionality is to have a "message type" field in the Message.

libqaul/common/src/message.rs
```diff
@@ -0,0 +1,16 @@
...
+
+/// A `Message` to be sent from the `sender` to the `recipient`, associated
+/// with the given `service`.
+/// Differs from a RATMAN message; this is a user-to-user, fully encapsulated
+/// message which does not care (or know) how it is delivered, merely that it
+/// was sent by one user and is meant for another.
+pub struct Message {
+    sender: UserID,
+    recipient: UserID,
+    // TODO: Add some kind of cryptographic signing here
```
spacekookie: I suggest we just add a new type for now, called `MsgSignature` (name up for :bike::house:), that contains a `u32` for a unix timestamp. We can later amend it with whatever format our signatures will have.

It's a bit of an open question. The old codebases used `mbedtls` but I haven't looked at the Rust crate for it yet and don't know how nice it is to work with. There's also a bunch of alternatives to work with. Plus, I kinda got it into my head that maybe we could implement the Axolotl ratchet.

**TLDR:** Maybe add that `MsgSignature` type into that file, add it as a field, just to make sure we don't forget it. The rest can be seen later and we an merge this :)

libqaul/common/src/payload.rs
```diff
@@ -0,0 +1,68 @@
...
+/// no transmission errors occurred (message digest of payload type and data).
+#[derive(PartialEq, Eq, Debug)]
+pub struct Payload {
+    length: u64,
+    data: Vec<u8>,
+    digest: HashBytes
+}
+
+// Compute and return the 64-bit Blake2b digest of the given data.
+fn blake2b_digest(data: &[u8]) -> HashBytes { Blake2b::digest(data) }
```
spacekookie: Code-style wise we don't currently enforce the rule of `rustfmt` (we will at some point probably) but I'm kinda against "in-lining" function bodies like this. 

I've read too much C code and my brain just parses it as a header definition and then I get confused about what it is in Rust :sweat_smile:

NoraCodes: OK! I can just rustfmt everything prior committing, that might be best

libqaul/common/src/message.rs
NoraCodes: Will do!

spacekookie (approved):

PR #246, opened 2019-06-19
First approximation of the visn testing framework

NoraCodes:
`visn` provides a way to unify testing of eventually-consistent
components of the software, abstracting out the logic to queue up events
and replay them in various orders. New functionality will be added as
tests require it.


spacekookie (approved):
This looks really cool!

If you want we can merge this as-is. Although I'm wondering if maybe this library would make sense as a crate in it's own repo? (on the `qaul` organisation). Up to you though!

NoraCodes:
I can definitely pull it out into its own repo, if that's easier; I put it here mostly because I intend to use it pretty heavily and don't want to have to deal with constantly keeping versions synchronized.

spacekookie:
On 19-07-01 08:35, Leonora Tindall wrote:
> I can definitely pull it out into its own repo, if that's easier; I put it here mostly because I intend to use it pretty heavily and don't want to have to deal with constantly keeping versions synchronized.

Oh please, I really don't mind it being in this repo, especially in
the beginning!  Just wanted to let you know that if you preferred it
to be external, I'm very open to that as well.

PR #248, opened 2019-06-28
[Draft] Simple user initialisation

spacekookie:
This PR adds simple user initialisation to `libqaul`,
as well as required state in the `Qaul` context struct.

For now, we're gonna use `Arc<Mutex<T>>` because it's easy
to work with, but we might want to look at some lock-free
datastructures in the future instead.


spacekookie:
Some changes that are important here @Jess3Jane:

- I added a `pw` field to both the user creation and login functions. That's obviously important to take a user secret and do $stuff with it. And the lack of this was probably what was confusing you in the other PR about how authentication would work
- I derived some stuff  for`Identity` to make it easier to work with

I think I might merge this PR pretty soon, so that you're not blocked anymore. Then next I'll work on the auth module which will handle stuff like verifying user passphrases and keep track of tokens and stuff. I think at that point imagining the interactions of the API are gonna feel a lot more natural.

PR #249, opened 2019-06-29
Http authenticator

Jess3Jane:
a wip pull request, more of a request for comments if you will. i'm unsure on the following and would like input, as well as on anything else:

**Error Struct**
presently there's an error struct in the `src/api` area but that appears to be for service errors? this request has an error struct for internal libqaul errors but i'm not sure this is the best way. other options include:

- returning a module level error struct
- returning the underlying error struct directly for structs which have only one kind of error
- decomposing internal errors into a shared error struct

**Location of http-api**
presently it's it's own little crate but with how much of it will have to be integrated with core and the service api does it make more sense to have it be in `src`?


Jess3Jane:
another question, should the `ApiServer` struct actually take an instance of Qaul (via whatever method) or should it take like
an authentication function and such so it doesn't have a reference to its parent to prevent accidental deadlocks

spacekookie:
Haven't looked at the code yet (will do that in a second), but just some stuff that'll be relevant that I'm also working on in #248:

if we look at the structure of a qaul.net application, there's `libqaul` providing a service API and `api/error` is pretty much errors around that API. But the HTTP API is provided for two or three specific services (`messaging`, `filesharing`, `voice`). *Maybe* those errors will line up somewhat with the underlying service layer errors, but I wouldn't want to couple them too tightly.

What I'd suggest we do is you just mirror errors in your layer for now, until we have the actual main services in place, and if we realise that something can be streamlined (i.e. code deleted), we can still do that then.
But it means that you're more flexible now.

Also, regarding instances of `Qaul`: every service function is `&self`, and mutability is handled via internal locks (`Arc<Mutex<T>>` for now). That means we can be very liberal about just passing around borrowed references to stuff. 
Although I'm wondering now, how much sense it makes to give the HTTP ApiServer access to the `Qaul` instance and not, say, just the instances of the services. Those don't exist yet, so I guess I'll spend this train ride writing those :sweat_smile: 

Anyway, I'll have a look at code in a second, just wanted to clarify those things

libqaul/Cargo.toml
```diff
@@ -1,13 +1,14 @@
 [package]
 name = "libqaul"
 version = "0.1.0"
-authors = ["Katharina Fey <kookie@spacekookie.de>"]
+authors = ["Katharina Fey <kookie@spacekookie.de>",
+		   "Jess 3Jane <me@jess.coffee>"]
 edition = "2018"
 
 [dependencies]
+iron = "0.6"
```
spacekookie: This looks like it should only be a dependency of `http-api` ?

libqaul/src/auth/mod.rs
```diff
@@ -1 +1,12 @@
+pub struct Authenticator;
+use crate::{
+    error::QaulResult,
+    users::User,
+    Qaul,
+};
 
+impl Qaul {
+    pub fn authenticate(&self, token: &str) -> QaulResult<Option<User>> {
```
spacekookie: Have a look at the API that I added in #247 (i.e. maybe rebase `master` :stuck_out_tongue:). There's `user_login` and `user_logout`, plus the concept of an `AuthUser` (i.e. something that wraps both the `Identity` and `Token` of a session).

Something that function currently lacks (which...feel free to add it!) is a `secret` argument, i.e. a password.

spacekookie (commented):
Left some comments. But otherwise it looks good :+1:

spacekookie:
Oh, I also did #251 if you wanna have a look at the general structure of the `messaging` and `files` service

(also not sure about `files` vs `filesharing` :thinking:)

libqaul/Cargo.toml
Jess3Jane: presently its a dependency here because error implements `From<HttpError>` and `From<Error> for IronError`

libqaul/src/auth/mod.rs
Jess3Jane: Hmmm that won't quite work in this case as we don't have the identity of the user we're logging in, just an api key

with this when you start an application you do normal first party login to get the token, which is certainly doable, but also seems somewhat of a pain (and probably encourages applications to store the secret to prevent the user from having to log in every time)

Jess3Jane: or is the idea that that part is stored in the api layer?
i know there's some concept of logging in/logging out for the crypto things y'all are doing so it'd be a world of several keys regardless

Jess3Jane:
having issues getting the lifetimes to work out for taking `&Qaul`
changed it to take `Arc<Qaul>` for now 

rustc ends up convinced anything in the iron typemap needs to outlive `'static`

spacekookie:
On 19-06-29 06:57, Jess wrote:
> or is the idea that that part is stored in the api layer?
> i know there's some concept of logging in/logging out for the crypto things y'all are doing so it'd be a world of several keys regardless

Hmm. Generally the interaction should be

1. App (service) signs-in user
2. App get's session token
3. That token can be re-used as long as the underlying instance
   doesn't change

The token should be cached by the app, not `libqaul`. If a user
insists on having their login cached, that by design reduces security
but I feel that it's okay to then expect the service API/ the actual
service app to cache this session token.

spacekookie:
> presently its a dependency here because error implements `From<HttpError>` and `From<Error> for IronError`

Ah, and I guess you can't implement that `From` otherwise because of
orphan rules? The alternative would be to just wrap the `QaulError` on
the API layer and use something like `failure`, which might be a good
idea regardless.

For now I'm kinda okay having this dependency here, although maybe put
it behind a feature flag. And we should look into removing the
dependency from the core lib in the future.

Jess3Jane:
> For now I'm kinda okay having this dependency here, although maybe put it behind a feature flag. And we should look into removing the dependency from the core lib in the future.

ive moved it into the http-api crate so the dependency isnt there anymore

Jess3Jane:
> Hmm. Generally the interaction should be 1. App (service) signs-in user 2. App get's session token 3. That token can be re-used as long as the underlying instance doesn't change The token should be cached by the app, not `libqaul`. If a user insists on having their login cached, that by design reduces security but I feel that it's okay to then expect the service API/ the actual service app to cache this session token.

that all makes sense but presently i dont see any way to exchange a session token for an auth user or the like

Jess3Jane:
> that all makes sense but presently i dont see any way to exchange a session token for an auth user or the like

i think I'm likely misunderstanding

so if I'm an http client application communicating with a service, what information do i need to provide such that qaul can authenticate i am allowed to do so and so that the service can know who i am

or should that be handled in the http api?

spacekookie:
> i think I'm likely misunderstanding
> 
> so if I'm an http client application communicating with a service, what information do i need to provide such that qaul can authenticate i am allowed to do so and so that the service can know who i am
> 
> or should that be handled in the http api?

Yea, I think there's some misunderstanding here 😅

The information you need to provide is the "User ID" (more to that in a
second) and the user secret (passphrase or whatever). `libqaul`
authenticates that, then returns you (UserID, Token), where that pair
is then used for all future operations.

I put "User ID" in quotes and the API is returning the ID explicitly
because something we might want to do is abstract the combersome
fingerprint ID with a cleartext username, that the user get's to
choose (there's already different "name" fields in the User structs).

In that case, the user would provide something like "spacekookie" and
"changeme" and the service API returns "<fingerprint>" and "<token>".
The token should then be cached by the application and re-used for the
session (although `libqaul` will probably make them time-out at some
point.

Does that make sense?

Jess3Jane:
yes! okay that makes sense

I'll update this pull request later with token caching

Jess3Jane:
i removed `error.rs` because on further consideration it's probably no how i want to handle errors
the authenticator struct now `HashMap`s a `String` token into an `Identity`
i also added `Copy` and `Clone` to `Identity` as it's just wrapping a vec and there's no easy way to turn `&Identity` into `Identity` otherwise

i was going to write some unit tests but `iron-test` is maybe not as usable as i was hoping so i'm evaluating options

Jess3Jane:
as you may notice there is currently no way to do user login/logout

this will come next

Jess3Jane:
> this will come next

i suppose i should mention, this will be a separate pull request

spacekookie:
I'd be happy merging this for now. Can you please rebase against `master`? There's a few commits in there that shouldn't be (which is probably why there's a conflict)

(`rebase` as opposed to just merging master in. It makes for a slightly cleaner git history in the end because I don't want to squash the rest)

Jess3Jane:
i gave up at rebasing and just applied the changes over `master` and submitted another pull request over at #253 

i'll try to be a bit less messy with my git commits next time to make this easier

PR #251, opened 2019-06-29
qaul.net service drafts

spacekookie:
This PR adds two services: messaging and files,
which allow people on a qaul network to exchange
text messages (like e-mail) as well as file-share links.

The API of the `messaging` service is flexible enough
to also allow inlining if a payload isn't too big.
Alternatively, the fallback "share link" will be used,
which mirrors the current behaviour of qaul.net (`<2.0.0`)

### DIFF
--- a/libqaul/service/files/src/lib.rs
+++ b/libqaul/service/files/src/lib.rs
@@ -1,7 +1,32 @@
-#[cfg(test)]
-mod tests {
-    #[test]
-    fn it_works() {
-        assert_eq!(2 + 2, 4);
+//! `qaul.net` filesharing service
+
+use qaul::{Qaul, QaulResult, UserAuth};
+use identity::Identity;
+
+/// A typed file that can be sent across the network
+pub struct File {
+    mime: String,
+    data: Vec<u8>,
+}
+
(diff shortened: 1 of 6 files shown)


libqaul/service/files/src/lib.rs
```diff
@@ -1,7 +1,32 @@
...
-    fn it_works() {
-        assert_eq!(2 + 2, 4);
+//! `qaul.net` filesharing service
+
+use qaul::{Qaul, QaulResult, UserAuth};
+use identity::Identity;
+
+/// A typed file that can be sent across the network
+pub struct File {
+    mime: String,
```
Jess3Jane: might make sense to use the hyperium `mime` crate?
https://github.com/hyperium/mime

spacekookie:
On 19-06-29 08:51, Jess wrote:
> might make sense to use the hyperium `mime` crate?
> https://github.com/hyperium/mime

Oh yea, I've had a look at that crate before and it's probably a good
idea to use it!

spacekookie:
I'm just gonna go ahead and merge this. All further development can happen in other PRs

PR #252, opened 2019-07-02
Add validation for frame payloads with BLAKE2b

NoraCodes:
Move all data (but not metadata) in the Frame into a Payload which has a utility function for validating its length and digest.

The only real question with this PR is, should we also validate metadata (an FCS, like Ethernet and some other wire protocols use)? This will make version interoperability much more difficult; if we add a field or change the size of a field, mismatched versions will be unable to validate each others' frames, unless we add an additional field to keep track of the version.

On the other hand, errors in e.g. the sequence number could be quite problematic for transmission and reassembly.


spacekookie (approved):
Looks good to me :)

PR #253, opened 2019-07-03
Added http-api with authenticator

Jess3Jane:
- Added http-api
  - Added authenticator middeware to http-api
  - Removed old qaul-http api
  - Changed identity to be Copy/Clone
  - Added user stub
  - Cleaned up a few imports


Cargo.toml
```diff
@@ -11,6 +11,9 @@ members = [
     "libqaul/service/files",
     "libqaul/service/core",
 
+	# qaul.net http-api
+	"libqaul/http-api",
```
spacekookie: Can you please use spaces for indentation

libqaul/http-api/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "http-api"
```
spacekookie: The crate name should probably be more descriptive like `qaul-http` :sweat_smile:

libqaul/http-api/src/lib.rs
```diff
@@ -0,0 +1,78 @@
...
+    }
+}
+
+struct QaulCore{
+    qaul: Arc<Qaul>,
+}
+
+impl QaulCore {
+    pub fn new(qaul: Arc<Qaul>) -> QaulCore {
+        QaulCore{
```
spacekookie: Formatting is a bit weird here. Also maybe just use `Self`

spacekookie (approved):
There's some small formatting changes, but otherwise this looks very good.

I assume that part of what the authenticator does here will be able to be moved to the `auth` module in `libqaul` at some point, but having it here for now is totally cool to get started

libqaul/http-api/src/lib.rs
Jess3Jane: I changed things to use `Self` and reformatted a bit, if you have any other formatting concerns let me know

Jess3Jane:
I also added a commit that derives `Clone` on `Qaul` to allow references to be shared across thread boundaries as discussed on the Zulip

I'm still not overly sold on this way of doing it but we need some way to share `&Qaul` across thread boundaries

spacekookie:
Yea, agreed. It feels like more of a hack for now. What we could possibly do is just do a type alias for `Qaul = Arc<ActualQaul>` or something. But we can bikeshed this later too.

PR #254, opened 2019-07-04
Implement and test user functions in the Qaul API

NoraCodes:
Just wanted to put this up to get feedback. It shouldn't be merged until I add some tests. See commit messages for more info.


libqaul/http-api/src/auth.rs
```diff
@@ -1,23 +1,17 @@
 use crate::QaulCore;
 use iron::{
-    BeforeMiddleware,
```
spacekookie: Could you maybe revert your `rustfmt` commit and only re-run it on files that you touched? That'll make the PR a bit more clear.

We should definitely pin some `fmt` task in CI to prevent accidentally merging things that aren't properly formatted (whoops :speak_no_evil:)

libqaul/src/api/models.rs
```diff
@@ -28,6 +27,16 @@ pub enum UserAuth {
     Trusted(Identity, String),
 }
 
+impl UserAuth {
+    /// Returns an error if the UserAuth isn't Trusted.
+    pub fn require_trusted(self) -> QaulResult<(Identity, String)> {
```
spacekookie: That's a nice util function! :+1: 

Although maybe make it `pub(crate)` and shorten the name to something like `trusted` ? (just as an idea)

libqaul/src/users/mod.rs
```diff
@@ -4,14 +4,35 @@ use identity::Identity;
...
-    pub(crate) bio: BTreeMap<String, String>,
-    pub(crate) services: Vec<String>,
-    pub(crate) avatar: Option<Vec<u8>>,
+/// A complete user, with ID and data.
+#[derive(Debug, PartialEq, Clone)]
+pub struct User {
+    /// A users network ID
+    pub id: Identity,
+    /// The user's information, such as name, bio, etc.
+    pub data: UserData,
```
spacekookie, line 13: This is an interesting change. Initially I wanted to split the representation of a user `libqaul` internally and service API externally because internally we will probably also end up storing some secret stuff that we definitely don't want to accidentally leak outwards.

On the other had, we could probably just keep that state elsewhere and refer to it via the user id or something.

All in all, this feels more elegant, so :+1:

spacekookie (commented):
Cool! I left a few comments, although some of them are just "cool" :sweat_smile: 

Also note the thing about `rustfmt`, I'm gonna sit down and fix CI to include some checks with a pinned `fmt` or something

libqaul/src/api/models.rs
NoraCodes: Yeah, good call.

libqaul/src/api/mod.rs
```diff
@@ -35,26 +37,48 @@ impl Qaul {
...
     /// to encrypt
     pub fn user_create(&self, _pw: &str) -> QaulResult<UserAuth> {
-        let user = _User::new();
+        let user = User::new();
         let id = user.id.clone();
         let mut users = self.users.lock().unwrap();
         users.insert(id.clone(), user);
 
-        Ok(UserAuth::Default(id))
+        Ok(UserAuth::Trusted(id, "fixme".into())) // TODO: Actually generate passphrase
```
spacekookie: I'd rather this line was a call to an `unimplemented!()` function in the `auth` module. We should let unimplemented things crash, rather than have "fixme" values :sweat_smile:

libqaul/src/lib.rs
```diff
@@ -2,19 +2,19 @@
 
 mod auth;
 mod crypto;
-mod users;
+pub mod users;
```
spacekookie: Actually I think maybe we keep the module private, then use fascading to expose only what people need (i.e. `User` and `UserData`. Considering there'll be more stuff in `users`, this feels more elegant and we can have a simple `api` prelude or something)

libqaul/src/lib.rs
```diff
@@ -2,19 +2,19 @@
...
 pub use api::*;
 
-use identity::Identity;
+pub use identity::Identity;
 use std::{
     collections::BTreeMap,
     sync::{Arc, Mutex},
 };
-use users::User;
+use users::{User};
```
spacekookie: I guess this used to be more than one import? :D

spacekookie (approved):
This is looking really really cool! I have a few nitpicky comments about style and export visibility but otherwise, it's great! :ok_hand: 

Happy to merge this after a few small changes.

libqaul/src/lib.rs
NoraCodes: Ah yeah, guess I forgot to run `rustfmt` for that last commit.

libqaul/src/lib.rs
NoraCodes: Definitely. I'll expose only those structs here.

libqaul/src/api/mod.rs
NoraCodes: I chose not to do that because it would cause every test to autofail - if that's what we want, maybe I can make the unimplemented bit `#![cfg(not(test))]`? That feels wrong, somehow.

NoraCodes: Oh, or, another option: we could have some way of injecting a user with the desired key material into the `Qaul` directly. I'll prototype that and see how well it works.

NoraCodes: This is done in a714d91 and c689a18

libqaul/src/lib.rs
```diff
@@ -2,19 +2,20 @@
...
 
+mod users;
+pub use users::{User, UserData};
+
 // This module defines the libqaul service API
 mod api;
 pub use api::*;
 
-use identity::Identity;
+pub use identity::Identity;
```
spacekookie, line 14: I just realised this is the exact same change as #257, so let's just resolve the conversation over there by merging this PR and we can take another stab at the API visibilities later :sweat_smile:

visn/src/lib.rs
```diff
@@ -57,71 +57,127 @@ use std::collections::VecDeque;
...
 ///
-
-pub trait KnowledgeEngine<S, E>: Sized {
-    fn queue_event(self, event: E) -> Self;
-    fn queue_events(self, events: &[E]) -> Self;
-    fn resolve_with<F: FnOnce(&mut dyn Iterator<Item = E>) -> &mut dyn Iterator<Item = E>>(
+/// # Types
+/// - `System`: the type of the system under test.
+/// - `Event`: the type of synthetic events.
+/// - `Return`: the type returned by the `resolve` function. Can be the same as `S`, or
```
spacekookie: Still referencing `S` here.

Btw, I'm not against using not-single letter type names here, but am curious why you wanted to spell them out. `S`, `E` and `R` wouldn't have been _too_ obscure, probably

spacekookie (commented):
Just some minor comments and then we'll merge this. Also I guess closing #257 because of this

visn/src/lib.rs
NoraCodes: I personally found them really confusing while doing refactorings but that might not impact the end user as much, I'm not sure.

NoraCodes: To expand on that: `S`, `E`, `R` aren't to bad by themselves, but with `F` and `G` for function types and `P` for the concrete internal success type for `FallibleEngineImpl`, it gets confusing.

spacekookie (approved):
Excellent, looks great now!

PR #257, opened 2019-07-20
added pub to libqaul/identity

MathJud:
the http-api needs identity to be pub, otherwise it doesn't compile
needed to build rust documentation

### DIFF
--- a/libqaul/src/lib.rs
+++ b/libqaul/src/lib.rs
@@ -9,7 +9,7 @@ mod storage;
 mod api;
 pub use api::*;
 
-use identity::Identity;
+pub use identity::Identity;
 use std::{
     collections::BTreeMap,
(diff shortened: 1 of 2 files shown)


spacekookie (commented):
We can do this (essentially re-exporting `Identity`).

Alternatively we could also have the rest API include the `ratman-identity` crate directly.

I'm not sure about the pros and cons of this.

MathJud:
I actually don't have an opinion on that!
The only thing I needed was the documentation to compile. 
It failed due to the /http-api/ using libqaul.Identity

I just added a script folder with the build and upload-script for the qaul-rust-documentation.
It is accessable under: https://api.qaul.net

spacekookie:
Closing in favour of #254

PR #272, opened 2019-07-22
Add partial user updates (deltas)

NoraCodes:
This will reduce the size of user updates, and provides a way to blank previously set UserData values.


NoraCodes:
Ping @spacekookie: does this look OK? Some of my other stuff is currently based on this and I'd like to either merge it or redo it without this.

spacekookie (approved):
Oof, yea sorry I thought I had already given you feedback to this. Looks good to me!

PR #275, opened 2019-07-24
Add login and logout endpoints

Jess3Jane:
This should probably not be merged until i document it a little better but as the title says there's now actual endpoints on the api

things to note:
`qaul-http::models` contains the JSON:API models used by the endpoints
authenticator stuff has been moved into `qaul-http::auth`
so much code is required to produce meaningful errors that i've split that out into a different file as it makes the main flow of `login` and `logout` much cleaner


libqaul/http-api/src/auth/authenticator.rs
```diff
@@ -0,0 +1,52 @@
...
+        }
+    }
+}
+
+impl typemap::Key for Authenticator { type Value = Self; }
+
+impl BeforeMiddleware for Authenticator {
+    fn before(&self, req: &mut Request) -> IronResult<()> {
+        if let Some(bearer) = req.headers.get::<Authorization<Bearer>>() {
+            if let Some(identity) = self.tokens.lock().unwrap().get(&bearer.token) {
```
spacekookie: It might be nicer to rewrite this block in a more functional style, i.e. with `.get::<...>().and_then(|bearer| ...).and_then(|identity| ...)`, etc.

libqaul/http-api/src/auth/error.rs
```diff
@@ -0,0 +1,137 @@
...
+            AuthError::QaulError(QaulError::UnknownUser) => Some("Unknown User".into()),
+            AuthError::QaulError(QaulError::InvalidQuery) => Some("Invalid Query".into()),
+            AuthError::QaulError(QaulError::InvalidPayload) => Some("Invalid Payload".into()),
+            AuthError::QaulError(QaulError::CallbackTimeout) => None,
+            AuthError::NotLoggedIn => Some("Not Logged In".into()),
+        };
+
+        let detail = match self {
+            AuthError::ConversionError(ObjectConversionError::ImproperType{ expected, got }) => 
+                Some(format!("Primary data should be of type {} but is of type {} instead",
```
spacekookie, line 75: Not something that's really relevant for this PR, but maybe we should keep an eye on is translation services. With this much detail in the error messages, it'd be kinda cool if we could pull in translated strings somehow and let service authors (or us) be able to use some of them to directly display in the presentation layer.

...maybe

libqaul/http-api/src/auth/login.rs
```diff
@@ -0,0 +1,104 @@
...
+            return Err(AuthError::NoSecret.into());
+        },
+    };
+
+    let qaul = req.extensions.get::<QaulCore>().unwrap();
+
+    // perform the login
+    let (ident, token) = match qaul.user_login(identity.clone(), &secret) {
+        Ok(QaulUserAuth::Trusted(ident, token)) => (ident, token),
+        Ok(QaulUserAuth::Default(_)) => {
```
spacekookie: Maybe just put the `unreachable!()` on the next line. Maybe we should also rename the `UserAuth` variant here to be more self descriptive that makes the comment obsolete

libqaul/http-api/src/auth/login.rs
```diff
@@ -0,0 +1,104 @@
...
+    // register the token with the authenticator
+    {
+        req.extensions.get::<Authenticator>().unwrap()
+            .tokens.lock().unwrap()
+            .insert(token.clone(), ident);
+    }
+
+    // return the grant
+    // so you'd think the id here would be the user id right?
+    // NO
```
spacekookie: Oof

spacekookie (commented):
Hey, this is looking pretty good!
I haven't gotten all the way through the PR yet, but I left you some notes. I'll try to get the rest of it later. 

Also, could you write a general short overview of what's still missing from this PR?

libqaul/http-api/src/auth/error.rs
Jess3Jane: that'd definitely be a good idea and probably falls under the "configuration management" area of my proposal so I'll definately have a think about it

libqaul/http-api/src/auth/login.rs
Jess3Jane: yeah... in  this case i might make it just the token type and return a success message for cookie grants but in general this is a problem and this is roughly the community standard solution

which is obviously still kinda weird but >.>

Jess3Jane:
The major thing i need to work on is documenting this
there's a bit of the authenticator stuff that will be needed by any services wishing to present an http api and those should get doc comments
i also need to document endpoints and entities but i'm not really sure where to do that?
finally i'd like to add another grant type to allow for granting server side cookies instead of returning a token

libqaul/http-api/src/auth/authenticator.rs
Jess3Jane: ah i was having issues getting ownership to work out but double `and_then` worked thankyou

libqaul/http-api/src/auth/login.rs
Jess3Jane: i renamed it to `UserAuth::Untrusted` which hopefully better reflects why that is a meaningless return

Jess3Jane:
added the grant type but login cookie will be a separate pr later

Jess3Jane:
:heavy_check_mark: document public structures
:question: i have documented the http-api in its own book at the moment but i'm not sure that's the best place for it

docs/contributers/src/SUMMARY.md
```diff
@@ -9,10 +9,12 @@
...
+		- [HTTP](./technical/api/http/_intro.md)
     - [`libqaul`](./technical/libqaul/_intro.md)
     - [Web GUI](./technical/webgui/_intro.md)
 - [Translations](./translations/_intro.md)
 - [Design](./design/_intro.md)
 - [Legal](./legal/_intro.md)
     - [GPL License (Version 3)](./legal/gpl-3.0.md)
-    - [Creative Commons Attribution 4.0 International License](./legal/cc-by.md)
\ No newline at end of file
+    - [Creative Commons Attribution 4.0 International License](./legal/cc-by.md)
```
spacekookie, line 20: What happened with that line? :sweat_smile:

docs/http-api/src/entities/user.md
```diff
@@ -0,0 +1,19 @@
+# `user`
+A qaul.net user
+
+## Id
+The user's identity encoded in Base64 with the URL-safe character set.
+
+## Attributes
+### `display_name: String` _(optional)_
+A human readable display-name (like @foobar)
```
spacekookie: Maybe put "@foobar" into `code` blocks

docs/http-api/src/topics/authentication.md
```diff
@@ -0,0 +1,45 @@
...
+{
+	"data": {
+		"id": "0000000000000000",
+		"type": "user_auth",
+		"attributes": {
+			"secret": "my super secret password",
+			"grant_type": "token",
+		}
+	}
+}
```
spacekookie, line 25: This code block seems to be indented with a mix of tabs and spaces

spacekookie (commented):
I skimmed through the documentation you added and it looks good at a glance. I made a few comments about formatting, but those are all pretty minor.

Also, I kinda feel like putting the `http` docs in the `contributers` guide is gonna be nicer. It means people will only need to get a single book, it'll be easier to host, and also it just adds more context to everything people are looking at.

Not sure if it's possible to make `mdbook` fold away items until they're being expanded, but even if not, I think putting it all in the same book might still be nicer

docs/contributers/src/SUMMARY.md
Jess3Jane: ?-? i haven't a clue

docs/http-api/src/topics/authentication.md
Jess3Jane: That's very strange as it's not on my local copy

Jess3Jane:
Unfortunately there does not currently seem to be a way to fold interior chapters, although i agree that would be ideal
https://github.com/rust-lang-nursery/mdBook/issues/187

Jess3Jane:
i have reformatted the authenticator again because it should probably not continue as if nothing is wrong when a request has an invalid token

with that other than moving the documentation which i'm not sure if we want to do i think that's all i have for this pr so i'll mark it as ready

spacekookie (approved):
I'll just merge this as is for now. We can yakshave the documentation issue later. I think having it all i a single book would be good. Maybe we can patch mdbook to make that happen. But that's beyond the scope of this PR.

Otherwise this looks really good and thank you for all your effort! :heart:

PR #276, opened 2019-07-25
Implement contacts_find and contacts_get_all

NoraCodes:


libqaul/src/api/mod.rs
```diff
@@ -126,13 +126,29 @@ impl Qaul {
...
+            .iter()
+            .filter(|(_, user)| user.data.like_query(&query))
+            .map(|(_, user)| user.clone())
+            .collect())
     }
 
     /// Enumerate all contacts known by a user
-    pub fn contacts_get_all(&self, user: UserAuth) -> QaulResult<Vec<User>> {
-        unimplemented!()
+    pub fn contacts_get_all(&self, _user: UserAuth) -> QaulResult<Vec<User>> {
```
spacekookie: This is slightly wrong behaviour.

While there is a list of users on a system, the contact book is a concept local to a user, and should only be handed out over passing authentication too

libqaul/src/api/mod.rs
```diff
@@ -126,13 +126,29 @@ impl Qaul {
...
+    pub fn contacts_find<S: Into<String>>(
+        &self,
+        _user: UserAuth,
+        query: S,
+    ) -> QaulResult<Vec<User>> {
+        let users = self.users.lock().unwrap();
+        let query = query.into();
+        Ok(users
+            .iter()
+            .filter(|(_, user)| user.data.like_query(&query))
```
spacekookie: Same thing here. Actually what we need first is a contact book store concept inside the user, before these endpoints can be implemented

spacekookie (changes requested):
I think maybe remove the actual end-point implementations for now, because they'll require a bunch of groundwork to happen first.

I do like the simple search function and the testing for it

spacekookie (approved):
We should sit down and figure out more libqaul structure soon

PR #322, opened 2022-03-02
feat: add search user functionality to new chat window

brenodt:


MathJud:
looks and works good!
would be nice to have a magnifying glass icon in the search line

PR #331, opened 2022-03-22
[fix] dynamic network view updates

brenodt:


MathJud:
I think the labels should be labeled the following:

* Bluetooth
* LAN
* Internet
* Show Entire Network

PR #338, opened 2022-04-12
fix: notifications feedback

brenodt:


MathJud:
Finer notification logic update

PR #347, opened 2022-06-06
[feat] improve Rust/Flutter cd pipelines

brenodt:


docs/flutter-release-template.md
```diff
@@ -0,0 +1,9 @@
+## Desktop Installers :desktop_computer:
```
MathJud, line 1: I like the Icons! :)

qaul_ui/installers/windows/inno_setup_win_installer.iss
```diff
@@ -0,0 +1,56 @@
+; Script generated by the Inno Setup Script Wizard.
+; SEE THE DOCUMENTATION FOR DETAILS ON CREATING INNO SETUP SCRIPT FILES!
+
+#define MyAppName "qaul.net"
+#define MyAppVersion "2.0.0"
+#define MyAppPublisher "Qaul"
```
MathJud: please write qaul in lower case letters.

Question:
Is it possible to change the MyAppVersion in the script dynamically, e.g. from the git tag?

MathJud (changes requested):
Thanks for this comprehensive update!
It looks great.

I only have a few points:

* I think it would make sense to put the installers folder out of the qaul_ui folder.
  * How about the utilities/installers folder?
* The icons should also go into an own folder, as they can be used by 
  * How about utilities/icons ?

What are your thoughts on this?

PR #369, opened 2022-07-18
[fix] Flutter: malformed UTF-16 String error

brenodt:


MathJud:
When only having an emoticon as a qaul user name, there is an error.
(e.g. 🌞)

Error log:

```
Another exception was thrown: Invalid argument (Name): not enough characters to form initials string: ""
flutter: [INFO] LogStorageManager	(2022-07-18 18:54:27.826653): storing log "qaul_log-ArgumentError-1658163267825.gzip" in directory:
flutter: 	$HOME/.local/share/qaul/Logs

```

Screenshot from 2022-07-18 18-51-04

PR #385, opened 2022-08-09
Common msg

MathJud:


MathJud:
- [ ] Update README.md of the CLI client

### group_net.proto

- [ ] Rename the Group Messages
  * invite_request
  * invite_response
- GroupNotify
  - [ ] There should be a list with user information of all users that joined the group newly.
    * for first joiners, there should be the entier users list of all group members. Why? Because we need this to be able to 
- [ ]  rename GroupInfoResponse to GroupInfo and create a new GroupInfoResponse that contains the GroupInfo message.


### group_rpc.proto

* Group messages
  - [ ] whenever there is a 'request' message, there should be a 'response' message, that sais: success/error. otherwise the UI does not know whether the request was successful. The following response messages are missing: rename, invite, remove, reply invite, send
- [ ] GroupConversationRequest message is never used and can be removed


## Integrate Direct Chat into Groupchat

* Each conversation has a group_id and sends the messages to all its members.
* The direct chat is a group with only two users, both are administrators.
  * They cannot invite other users. 
* We need a boolean flag in the group to show whether it is a direct chat group or not.
* We should create predictable group_id's for direct chats.
  * 8 Bytes from each user ID set together, smaller ID first.

* chat 
  * all messages are common messages


## Common Message Update

* Make CommonMessage easier:
  * Enum about which service
  * data -> binary

## RPC ChatMessage update

* we could create an enum to show the `type` of the message, this would ease the handling of it, I guess.

```
// message type
  enum MessageType {
    // default value, when nothing is defined.
    // drop this message
    NONE = 0;
    // chat message
    CHAT = 1;
    // file message
    FILE = 2;
    // info message
    INFO = 3;
  }
```

* The message `status` should also be represented by an enum, this makes it more clear

```
  enum MessageStatus {
    // message not sent yet
    SENDING = 0;
    // message sent
    SENT = 1;
    // message received
    RECEIVED = 2;
    // all group members received the message successfully
    // this option is only used for groups with more then 2 members
    RECEIVED_BY_ALL = 3;
  }
```

* we need a new ChatInfo message as a purely informative message for the chat conversation. e.g. 

``` 
// the possible info status
Enum InfoType {
  // default value, undefined message
  // delete this message
  NONE = 0;
  // user joined group
  GROUP_JOINED = 1;
  // user left group
  GROUP_LEFT = 2;
}

// the info message
message InfoMessage {
  // message type
  InfoType type = 1;
  // user ID of user joined or left
  bytes UserId = 2;
}
```

Updated ChatMessage

*  remove `is_group` flag
* change to enums

```
// a single chat message
// this message contains all possible message types
message ChatMessage {
    // index
    uint64 index = 1;
    // id of the sending user
    bytes sender_id = 2;
    // message id or member id
    bytes message_id = 3;        
    // message status
    ChatStatus status = 4;
    // conversation id(group_id)
    bytes conversation_id = 6;
    // time when the message was sent
    uint64 sent_at = 7;
    // time when the message was received
    uint64 received_at = 8;
    // message type
    MessageType type = 9;
    // content of the message    
    bytes content = 10;
}
```

## Crypto

- [ ] activate Crypto again

## Questions

* Which module should handle the sending of the messages into the group?
  * group module?
  * chat/conversation module?
  * each module itself (chat, files, rpc)?

PR #386, opened 2022-08-12
[refactor] (Rust) update Android build script

brenodt:


MathJud:
Thanks for this advanced script!

I found the following two issues:

1) On my machine `$ANDROID_NDK_HOME` points to a symbolic link. This made the script fail, as it was checking for directories.
  * I added another check for symbolic links.
2) I got the following warnings

```
find: warning: you have specified the global option -maxdepth after the argument -type, but global options are not positional, i.e., -maxdepth affects tests specified before it as well as those specified after it.  Please specify global options before other arguments.
find: warning: you have specified the global option -mindepth after the argument -type, but global options are not positional, i.e., -mindepth affects tests specified before it as well as those specified after it.  Please specify global options before other arguments.
```
  * I therefore reordered the options as suggested, and the warnings disappeared.

PR #388, opened 2022-08-18
fix: [Desktop] Tapping on new chat updates UI & visible messages

brenodt:


MathJud:
Thanks for this fix!
The switching works, but the link into the chat from the user profile is broken:

* when the chat does not exist:  it either shows the loading button forever
* when the chat exists: it displays the chat full screen, like on mobile.

MathJud:
works nice! Thanks

PR #398, opened 2022-09-12
fix: #394 - UI behaviours

### LINKED ISSUE #394: UI Behaviours
## Chat Tablet Mode

When being in Tablet mode, with a window only as little that only one Layout column is visible, we cannot write a chat message anymore, because the remaining one layout column always shows the chat overview view.

## Chat
Linux:
When sending an image and then another one. I only saw the first one twice.
The text was correct though.
When sending a third one, the second one showed the right image and the third one showed the first image ...
When clicking on the image itself, the correct one was displayed

MacOS:
no images were displayed in the chat. the font of the text for images was smaller than the text in other chat bubbles.

## Feed Messenger
When sending a new feed message, the feed modal overlay does not close after the sending of the message.

## Various Changes

- [x] Change Menu Entry `Classic Network View` to `Routing Data Table`

brenodt:
resolves #394


MathJud:
Was successfully merged into main

PR #424, opened 2022-10-07
[perf] improve high cpu usage

brenodt:


MathJud:
Fixes high CPU usage and reduces memory usage

PR #437, opened 2022-10-20
ci: [Rust] Add qauld compilation job

brenodt:


MathJud:
I think it would be good to have qauld: 

* downloadable as binary (as we had before)
  * could we add macos & windows targets here?
* downloadable as debian installer package for x86_64 & raspberry ARM architecture

brenodt:
@MathJud should be done now. Could you please check it out and let me know if you spot any issues?

MathJud:
Looks nice! Thanks a lot!

PR #438, opened 2022-10-23
Update

Kofituo:
This pull request creates wrappers for `send_sys` and `receive_sys` methods.

These methods can simply be referenced in Kotlin files as shown in [android/libqaul/src/main/java/net/qaul/libqaul/libqaul.kt file](https://github.com/qaul/qaul.net/blob/62dd1867660ab28059dfa3eb94fda0278f15e82b/android/libqaul/src/main/java/net/qaul/libqaul/libqaul.kt)


Kofituo:
A little side note. One of the added crates work with nightly. If you're uncomfortable with nightly for some reason, I could modify it so it works with Rust stable

PR #439, opened 2022-10-24
fix: revert confinement and change grade of snapcraft.yaml

brenodt:
@MathJud I found out that the issue on the Flutter CD pipeline was in the following configuration:

* `qaul_ui/snap/snapcraft.yaml:6`: `confinement` was set to `classic`, which is not supported by the Flutter snap extension - reverted back to [strict](https://snapcraft.io/docs/snap-confinement)
* `qaul_ui/snap/snapcraft.yaml:8`: `grade` was set to `beta`, which is an (invalid option)[https://snapcraft.io/docs/snapcraft-yaml-reference] - changed it to `devel`

With these options the job succeeds.

### DIFF
--- a/.circleci/continue-config.yml
+++ b/.circleci/continue-config.yml
@@ -280,8 +280,7 @@ jobs:
             - install-flutter-deps
             - run:
                 command: |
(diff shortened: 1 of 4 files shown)


MathJud:
Thanks for your findings & the fix!

Issue #374, opened 2022-07-27
Update Mechanism


MathJud:
Create the foundations to create update mechanisms that break the storage or configuration scheme.

This shall be done by having a version number stored on the system that describes the version of the last run qaul application. The current executable can check if this version is compatible or if an update mechnism needs to be triggered.

## TODO

- [x] create configuration file that stores version number
- [x] create update module
  - [x] read the version configuration file before anything else

MathJud:
closed by #458

PR #458, opened 2022-11-08
Features: Upgrade mechanism & internet node alias

MathJud:
## Feature Upgrade Mechanism

Added upgrade mechanism, to be able to automatically migrate installations with breaking changes in the data base and/or in the configuration file to the new version.

## Feature Internet Node Alias

Added a descripting alias name to each Internet node, for that users can recognize each node easier.


MathJud:
fixed #374

Issue #502, opened 2022-12-09
Linux Snap App only shows blank Window


MathJud:
When starting the snap qaul app under linux, the app only shows a blank window.

The reason seems to be the location where flutter tries to save the log files.
When starting the snap from the command line `snap run qaul` the following error is displayed:

```txt
$ qaul
flutter: storing log in directory: /home/sojus/Documents/Logs
[ERROR:flutter/lib/ui/ui_dart_state.cc(198)] Unhandled Exception: FileSystemException: Cannot create file, path = '/home/sojus/Documents/Logs/qaul_log-ProviderException-1670593224821.gzip' (OS Error: Permission denied, errno = 13)
#0      _File.throwIfError (dart:io/file_impl.dart:635)
#1      _File.createSync (dart:io/file_impl.dart:273)
#2      StorageManager._storeCompressedLog (package:logger/src/storage_manager.dart:137)
<asynchronous suspension>
```

MathJud:
when running the latest beta 12 build, we still get only a blank window to open.

The following snap error log is written:

```txt
= Seccomp =
Time: Dez 14 23:12:53
Log: auid=1000 uid=1000 gid=1000 ses=4 subj=snap.qaul.qaul (enforce) pid=2094375 comm="qaul" exe="/snap/qaul/x1/bin/qaul" sig=0 arch=c000003e 203(sched_setaffinity) compat=0 ip=0x7f23bc0bdc7f code=0x50000
Syscall: sched_setaffinity
Suggestion:
* add 'process-control' to 'plugs'

= AppArmor =
Time: Dez 14 23:12:53
Log: apparmor="DENIED" operation="open" profile="snap.qaul.qaul" name="/proc/sys/vm/max_map_count" pid=2094375 comm="qaul" requested_mask="r" denied_mask="r" fsuid=1000 ouid=0

```

brenodt:
@MathJud Since we merged #513, I'm closing this for now 🎉

MathJud:
fixed by #513

PR #503, opened 2022-12-12
[#502] snap permission-denied error

brenodt:
## Description
Find a fix for the #502 issue.
* Added `home` into the list of `Plugs`
* Added if clause to attempt to expand the `SNAP_COMMON` env var inside the Flutter app, and use that to store the Logs.

### Reasoning
Essentially, this happens because of the application running on it's own sandbox and not having access to the home directory [1].
We can add it as a required permission by adding `home` into our list of plugs, but that doesn't give us access to read/write hidden files [2].
Instead, we may want to use snap-specific paths, such as `SNAP_COMMON` [3].

### Testing
@MathJud Would you be able to run locally? You should be able to run it by doing this:
```bash
# compile for linux in release mode
cd rust
cargo build --release

# bundle the application into snap file
cd ../qaul_ui
snapcraft

# install snap locally
# We add the --dangerous flag to the end, as the snap file won't be signed at this point
sudo snap install qaul.snap --dangerous
```

I couldn't run locally as I'm having issues bundling the snap package. If you're not able either, we can always try running a release pipeline and testing the output.

### Misc
A useful tool to debug snap files is the `snappy-debug` snap.
You should install and run the `snappy-debug` command in a second terminal while running qaul, as it will print out issues that may arise, such as things blocked by confinement and how to solve them.

#### References
[1] https://snapcraft.io/docs/home-interface
[2] https://snapcraft.io/docs/permission-requests
[3] https://snapcraft.io/docs/data-locations
[4] https://forum.snapcraft.io/t/use-of-home-and-network-plugs/2587

### DIFF
--- a/docs/qaul/flutter/linux.md
+++ b/docs/qaul/flutter/linux.md
@@ -2,8 +2,16 @@
 
 ## Install Prerequisits
 
+Make sure you have the following tools installed, that you are 
+going to need for the further installation: git, curl
+
+```sh
+# Debian & Ubuntu
+sudo apt install git curl
+```
+
 * [Install Rust](qaul/rust/rust-install.md)
-* [Install Flutter](qaul//flutterflutter-install.md)
+* [Install Flutter](qaul/flutter/flutter-install.md)
 
 ## Clone Repository
 
--- a/docs/qaul/rust/rust-install.md
+++ b/docs/qaul/rust/rust-install.md
@@ -7,12 +7,16 @@ on your computer.
 
 Easiest install is via rustup: <https://rustup.rs/>
 
+## Further Tools needed for Rust Compilation
 
-## On Linux
+Install the following additional tools:
 
-Install compiler tool chain for C code
+* compiler tool chain for C code
+* protobuf compiler: protoc
+
+### On Linux
 
 ```sh
 # Debian, Ubuntu, Mint
-sudo apt install build-essential
+sudo apt install build-essential protobuf-compiler
 ```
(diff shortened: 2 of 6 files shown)


MathJud:
@breno thanks for the suggestions and the references!

I checked the use of the snap folders on several software. 

Similar software as qaul uses the `common` folder to store the user data (e.g. Skype).
This is a good behaviour, as the downloaded files and the data should remain the same from one version to the next.
In libqaul we take care of the upgrading ourselves.

I think, all the user data of libqaul should go to the snap `common` folder.
This means, that flutter should start libqaul when running in snap with the correct path to the snap `common` folder.

libqaul has a `logs` folder too. It is written with all lower case letters.
If we are saving the logs to the same location we should probably rename one of them to not get confused. How about renaming the `Logs` folder to `ui_logs`?

To request the `home` right within snap makes most probably sense anyway, as we will need it to save the files for the users somewhere else?!

### TODO

* start libqaul with path to `common` folder when running snap app.
* rename `Logs` folder.

brenodt:
I need to pass in the starting path to the `start` method when running on Linux - specifically for SNAP.

brenodt:
@MathJud That sounds most sensible, indeed.

I have attempted to update the FFI identity of the `Start` function and pass in the `SNAP_COMMON` env var to it, in hopes of having libqaul started from that path.

We can try creating a small release to check if this works. If not, I'll try to find a way of bundling the snap package locally using docker, instead.

PR #505, opened 2022-12-15
[#502] fix snap permission-related errors

### LINKED ISSUE #502: Linux Snap App only shows blank Window

brenodt:
## Description
The goal here is to fix #502 by:
* Improve testing feedback loop on snap packages
* Have qaul snap fixed so it works properly

### DIFF
--- a/.circleci/images/flutter-snapcraft/Dockerfile
+++ b/.circleci/images/flutter-snapcraft/Dockerfile
@@ -1,0 +1,32 @@
+FROM --platform=linux/amd64 ubuntu:focal
+
+# Install Common Dependencies
+ARG DEBIAN_FRONTEND=noninteractive
+RUN apt update && apt install -y \
+    git ssh tar gzip ca-certificates \
+    curl unzip xz-utils zip libglu1-mesa \
+    apt-transport-https clang cmake ninja-build pkg-config libgtk-3-dev apt-utils openssl snapd
+
+# Set up new user
+RUN useradd -ms /bin/bash developer
+USER developer
+WORKDIR /home/developer
+
+# Download Flutter SDK
+ENV FLUTTER_VERSION 3.3.9
+ENV PATH "$PATH:/home/developer/flutter/bin"
+RUN curl -s https://raw.githubusercontent.com/brenodt/flutter_utilities/main/bin/flinstall | bash /dev/stdin "$FLUTTER_VERSION"
+
+# Run basic check to download Dark SDK
+RUN flutter doctor
+
+# Install Snapcraft
+RUN snap install snapcraft --classic
+RUN snapcraft --version
+
+# Install Flutter Snapcraft Plugin
+RUN sudo snap install flutter-snapcraft --classic
+
+# Clone project and set it as the entrypoint
+RUN git clone https://github.com/qaul/qaul.net.git
+ENTRYPOINT ["/qaul.net"]
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
 .PHONY: bundle-snapfile
 bundle-snapfile:
-	@docker run --rm --env SNAPCRAFT_LOGIN_FILE=$(echo "$SNAPCRAFT_STORE_CREDENTIALS") --volume $(shell pwd):"/qaul.net" --workdir "/qaul.net" snapcore/snapcraft:edge /bin/bash -c 'sleep 1; cd utilities/installers/linux && bash snapbuild'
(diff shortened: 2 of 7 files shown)


brenodt:
@MathJud Would you be able to test the bundling by running `make bundle-snapfile` in the root? I'm having problems with docker on my Ubuntu machine for some reason.

brenodt:
@MathJud In trying to find a fix for the security issues, I found out the following in [this doc](https://snapcraft.io/docs/flutter-applications):

### Security model

> The next section describes the level of confinement applied to your app.
> 
> confinement: strict
> 
> Snaps are containerised to ensure more predictable application behaviour and greater security. Unlike other container systems, the shape of this confinement can be changed through a set of interfaces. These are declarations that tell the system to give permission for a specific task, such as accessing a webcam or binding to a network port.
>
> **It’s best to start a snap with the confinement in warning mode, rather than strictly applied. This is indicated through the devmode keyword.** When a snap is in devmode, runtime confinement violations will be allowed but reported. These can be reviewed by running journalctl -xe.
> 
> Because devmode is only intended for development, snaps must be set to strict confinement before they can be published as “stable” in the Snap Store. Once an app is working well in devmode, you can review confinement violations, add appropriate interfaces, and switch to strict confinement (our example uses strict because we know it’s working as expected).

Due to the application requiring the "flutter-stable", we cannot use _classic_ confinement - only _strict_ or _debug_.

So a potential work-around to the problem we're having is using `devmode + journalctl` to find out all confinement violations - which in turn determines which security interfaces - or plugs - our app needs to function and only then move to the _strict_ confinement.

This has the side-effect of us being unable to publish on the release channel - which requires strict or classic.
Until we have it fixed, we can install the application with:
```bash
snap install --beta qaul --devmode

# add --dangerous if installing directly from the *.snap file 
snap install qaul-<Version>.snap --devmode --dangerous
```

brenodt:
#### The issue
It seems the problem is that `liblibqaul.so` is not available at the snap bundle.

#### Possible solution
In order to link it to the package, it seems we need to define a layout [1], which describes where to copy over files into the snap environment.

More than that, we also need to specify an installation step [2] to have the library copied over to the `$SNAP_DATA/bin/lib` folder, where it can then be loaded by Flutter (see `qaul_ui/packages/qaul_rpc/lib/src/libqaul/ffi.dart:30`).

#### References
1. https://snapcraft.io/docs/snap-layouts
2. https://snapcraft.io/docs/supported-snap-hooks

##### Related issues:
* https://forum.snapcraft.io/t/need-to-store-specific-files-in-snap-data-directory/11356
* https://forum.snapcraft.io/t/unable-to-load-shared-library-so-in-application-using-snap/21639
* https://forum.snapcraft.io/t/layouts-re-mapping-snap-directories/1471


##### Here's some more useful documentation:
* [debugging snaps](https://snapcraft.io/docs/debug-snaps)
* [snap Env Vars](https://snapcraft.io/docs/environment-variables)
* [parts lifecycle](https://snapcraft.io/docs/parts-lifecycle)

MathJud:
We also have a `Seccomp` violation :
https://snapcraft.io/docs/debug-snaps#heading--seccomp

```txt
= Seccomp =
Time: Dez 16 20:36:51
Log: auid=1000 uid=1000 gid=1000 ses=4 subj=snap.qaul.qaul (enforce) pid=3462685 comm="qaul" exe="/snap/qaul/x1/bin/qaul" sig=0 arch=c000003e 203(sched_setaffinity) compat=0 ip=0x7f86312a6c7f code=0x50000
Syscall: sched_setaffinity
Suggestion:
* add 'process-control' to 'plugs'
```

This log can be seen when running `sudo journalctl --output=short --follow --all | sudo snappy-debug`

Issue #520, opened 2023-01-24
Dockerizing qauld to enable self-hosting


brenodt:
Opening an issue so it's easier to track the status of this feature.

### Discussed in https://github.com/qaul/qaul.net/discussions/518

<div type='discussions-op-text'>

<sup>Originally posted by **huang12zheng** January 24, 2023</sup>
I found that without documentation and best practices for relay nodes,
Maybe we could add it somewhere?

This is an example that I wrote
https://github.com/huang12zheng/qaul.net/tree/dockerfile</div>

MathJud:
The feature has been documented here: 
<https://qaul.github.io/qaul.net/#/qaul/infra/self-hosting-qauld>

And committed via this pull request: #526 

Thanks @brenodt for doing so

MathJud:
A nice thing to have for the future, could be the possibility to define the "Port" and the node "Name" in a `docker-compose.env` file and make it accessible to the docker container via the `environment` settings in  `dock-compose.yml`.

* [Docker-Compose `.env`-file documentation](https://docs.docker.com/compose/compose-file/#env_file)
* [Docker-Compose `environment`  option documentation](https://docs.docker.com/compose/compose-file/#environment)
* [An example of another docker-compose project using the `.env`-file and the `environment` option](https://github.com/open-comm/apache-static-ssh)

Just in case, that it would be preferable to overwrite the startup command in the dockerfile itself we could do so in the docker-compose.yml `command` section. IMHO the environment option gives us most probably enough flexibility.

* [Docker-Compose `command` documentation](https://docs.docker.com/compose/compose-file/#command)
* [Example Project that uses the `command` option](https://github.com/open-comm/traefik/blob/main/docker-compose.yml)

brenodt:
I do agree, this sounds reasonable.
One issue could be the fact that the `name` and `port` flags only take effect the first time the image is spinned-up - subsequent ones take the `config.yaml` file.
The user can, of course, change it manually: we already describe in the docs how to map the configuration folder into the host.
But this renders the environment variable only useful on the first execution of the container, I believe.

taro-ball:
One solution is for qauld to check for the presence of env variable and override the value from config.yaml if the variable is present. Maybe write something to the log like "env variable found, overriding config port value to wxyz". Could we request a feature like that? This is the best practice for containerized apps see https://medium.com/hackernoon/how-to-dockerize-any-application-b60ad00e76da (#7). 

[...]

PR #526, opened 2023-02-04
[#520] dockerize qauld

brenodt:
## Description
The aim is to have a docker instance of qauld to facilitate self-hosting.

@MathJud I think I'm going to separate the pipeline to create new images on it's own PR. Let me know if you spot any issues.

### DIFF
--- a/docs/_sidebar.md
+++ b/docs/_sidebar.md
@@ -1,6 +1,8 @@
 * [Qaul](qaul/README.md)
     * [Commit Guidelines](qaul/commit-guidelines.md)
     * [Glossary](qaul/glossary.md)
+    * **Self-Hosting**:
+      * [Your own community node](qaul/infra/self-hosting-qauld.md)
     * **Tooling**
         * [Configure Lefthook](qaul/tooling/configure_lefthook.md)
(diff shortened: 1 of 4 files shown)


MathJud:
@brenodt Thanks a lot looks great!

A nice thing to have for the future, could be the possibility to define the "Port" and the node "Name" in a `docker-compose.env` file and make it accessible to the docker container via the `environment` settings in  `dock-compose.yml`.

* [Docker-Compose `.env`-file documentation](https://docs.docker.com/compose/compose-file/#env_file)
* [Docker-Compose `environment`  option documentation](https://docs.docker.com/compose/compose-file/#environment)
* [An example of another docker-compose project using the `.env`-file and the `environment` option](https://github.com/open-comm/apache-static-ssh)

Just in case, that it would be preferable to overwrite the startup command in the dockerfile itself we could do so in the docker-compose.yml `command` section. IMHO the environment option gives us most probably enough flexibility.

* [Docker-Compose `command` documentation](https://docs.docker.com/compose/compose-file/#command)
* [Example Project that uses the `command` option](https://github.com/open-comm/traefik/blob/main/docker-compose.yml)

PR #542, opened 2023-03-23
[#540] Chat Menu not updating

### LINKED ISSUE #540: Group Settings Menu not displayed on Desktop
On Linux, the `Group Settings` options is not displayed when clicking on the three dots menu in the messenger in a group room.

When clicking on the three dots, `show menu` overlay hint is displayed.

Screenshot from 2023-03-21 09-24-20

brenodt:
## Description
Fixes #540.
The issue occurred when the user had a direct chat open (which has no menu options, thus no menu to open), and then opened a group chat (new or not).
Closing the current open direct chat prior to opening a group chat would remedy the bug, since the chat screen state was built anew.
Likewise, opening a group chat from an already-opened group chat would remedy the bug, as the menu options are the same.

## Solution
Update the overflow menu options on every chat state change ensures this bug no longer occurs.
As an UX improvement, I've also removed the menu button (three vertical dots) from the direct chats, as currently no menu options are there.

### DIFF
--- a/qaul_ui/lib/screens/home/tabs/chat/widgets/chat.dart
+++ b/qaul_ui/lib/screens/home/tabs/chat/widgets/chat.dart
@@ -131,17 +131,14 @@ class _ChatScreenState extends ConsumerState<ChatScreen> {
   @override
   void didChangeDependencies() {
     super.didChangeDependencies();
-
-    var l10n = AppLocalizations.of(context)!;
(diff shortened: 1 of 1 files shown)


MathJud:
I was testing the PR on Linux.
After creating a new chat group, the chat group appeared in the overview part of the window, but was not opened for chatting directly.

PR #550, opened 2023-04-14
file slicing and encryption in thread

Harshil-Jani:
In this PR, The file slicing and encryption logic is moved into the async thread. 
But at the cost of error handling. So, We may in future need to find a way to handle errors inside threads.


Harshil-Jani:
Maybe for handling the errors, We may want to take support for the [channels](https://doc.rust-lang.org/std/sync/mpsc/fn.channel.html) ? 
`sender` : would have error logic inside the new thread.
`receiver` : would handle errors back into main thread.

Issue #547, opened 2023-04-11
Add IPv6 Interface to Swarm


MathJud:
Currently qaul can be configured to be connectable either via IPv4 or via IPv6. In order for a libp2p swarm to become dual stack (IPv4 & IPv6), we need to add a second interface to each swarm.

The current interface starts listening to IPv4 and a second interface shall start listening to IPv6.

## Sub Tasks

- [x] Add a second listener to each libp2p swarm, which starts listening to incoming IPv6 connections.
  - [x] Add a second listener to the Internet connection. https://github.com/qaul/qaul.net/blob/899edace6dbd3e4e251f27b5498103a622b8afae/rust/libqaul/src/connections/internet.rs#L285
  - [x] Add a second interface to the LAN connection. https://github.com/qaul/qaul.net/blob/899edace6dbd3e4e251f27b5498103a622b8afae/rust/libqaul/src/connections/lan.rs#L256
- [x] Make interface listening address configurable. Via the configuration qaul configuration file.
  - https://github.com/qaul/qaul.net/blob/main/rust/libqaul/src/storage/configuration.rs

### Optional Advanced Tasks

- [x] Write an upgrade procedure, for clients to automatically upgrade to the new configuration file.

## Information

This is a good first issue, which touches quite a few areas of the 
Don't hesitate to ask further questions about it, e.g. in the matrix group chat. We will help you with that.

To have some further information for the fields that are touched here:

* **libp2p** - libqaul uses [rust-libp2p](https://docs.rs/libp2p/latest/libp2p/) crate for the LAN & Internet network connections. 
  * The [listen_on()](https://docs.rs/libp2p/latest/libp2p/trait.Transport.html#tymethod.listen_on) function reference.
* **Configuration** - the qaul configuration is written into a specific configuration file.
  * Documentation: http://qaul.github.io/qaul.net/#/qaul/rust/configuration
  * Internally libqaul uses the crate [config](https://crates.io/crates/config) for writing the configuration file.

## Testing

* Check node information. If it shows you the 
* You can test if the node is manually connectable by adding a static internet node to the

Harshil-Jani:
@MathJud In the `configuration.rs` do we need to use two structs for Lan and Internet (Eg : Lan_IPv4 and Lan_IPv6) ? Or there is already support for IPv4 connections through IPv4-mapped IPv6 addresses ?

If there is support already then do we really need two `Swarm::listen_on()` ? It should be possible with just one itself i guess.

[...]

PR #551, opened 2023-04-17
feat : Add IPv6 multiaddress

Harshil-Jani:
Solving #547 
- [x] Adding Listener to Internet connection 
- [x] Adding Listener to Lan connection
- [x] Make `listen` configurable with IPv4 and IPv6.
- [x] Automatic procedure for clients to upgrade to new configuration.


Harshil-Jani:
IPv4 and IPv6 MultiAddresses are now configured for both LAN and Internet. What we would do for writing an automatic upgrade procedure ?  I was reading some answers over here https://stackoverflow.com/questions/232347/how-should-i-implement-an-auto-updater and I think we would need to check if the configuration file now has IPv6 added to it or not and based on it we can run our script to update. 

Please let me know what could be a possible solution for this.

MathJud:
libqaul has an own [upgrade module in `libqaul/src/utilities/upgrade`](https://github.com/qaul/qaul.net/tree/main/rust/libqaul/src/utilities/upgrade), which is executed before loading the configuration, as a wrong file structure of the configuration results in a panic.
The decision of the upgrade procedure is bound to the library version.
We can bind it to the next version `beta 15`.

Each upgrade procedure is contained in an own folder. You can take the [upgrade to beta 9](https://github.com/qaul/qaul.net/tree/main/rust/libqaul/src/utilities/upgrade/v2_0_0_beta_9) as a reference.

MathJud:
The IPv6 works good, I just made the amount of interfaces per connection module dynamically configurable by using a for loop going through the configuration array.

PR #557, opened 2023-05-16
Fix issue with endless background task

brenodt:
## Description
On android, even if we close (kill) the application, the background task is not stopped. Apparently, there is no way to circumvent this behaviour, but I noticed that even without enabling it, I still see the android user as available when the app is running in the background.

> Note: I also took the opportunity to upgrade the gradle version being used, as I was having problems compiling for release.


brenodt:
Since the latest changes to the background task and our internal implementation, this PR is no longer useful.
Closing it.

PR #563, opened 2023-05-24
feat: Adding a relay bot for matrix<>qaul bridge

Harshil-Jani:
Raising the PR so we can do smooth code review for the code which we will be updating into the matrix-bridge branch.
As per the current state state
1. Run the bridge client `cargo run --bin=qaul-matrix-bridge`
2. Invite user `@qaul-bot` to any of the matrix rooms.
3. In message send `!qaul` and it will respond.


rust/clients/bridge/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "bridge"
```
MathJud: The name should be more explanatory, e.g. `qaul-bridge` or `qaul-matrix-bridge`. IMHO `qaul-matrix-bridge` would be best, as it is the most self-explanatory.

rust/clients/bridge/README.md
```diff
@@ -0,0 +1,86 @@
+# qaul-cli
```
MathJud: Please write a specific README for the binary, that explains what the binary does and how to start it.

rust/clients/bridge/src/relay_bot.rs
```diff
@@ -0,0 +1,78 @@
+use matrix_sdk::{
```
MathJud, line 13: Please add the license and a short description to every new file.
See e.g. `chat.rs` for an example.

rust/clients/bridge/src/relay_bot.rs
```diff
@@ -0,0 +1,78 @@
...
+    }
+}
+
+async fn login (
+    homeserver_url: &str,
+    username: &str,
+    password: &str,
+) -> Result<(), matrix_sdk::Error> {
+    // the location for `JsonStore` to save files to
+    let mut home = dirs::home_dir().expect("no home directory found");
```
MathJud: We should create a matrix specific `matrix.yml` config file, which is saved in qaul's default storage location, for that it is working as everything else in qaul.
The location to save the file to, can be asked from libqaul via the RPC debug>StoragePathRequest message. (see `rust/libqaul/src/rpc/debug.proto`)
If there is more matrix-bridge related stuff, i suggest to create an extra `matrix` folder in the qaul storage directory and use it as the storage directory for all matrix related stuff.
For saving and reading the configuration libqaul uses the crate `config`. (see `rust/libqaul/src/storage/configuration`.)

rust/clients/bridge/src/main.rs
```diff
@@ -0,0 +1,104 @@
...
+        std::thread::sleep(Duration::from_millis(10));
+    }
+
+    // initialize user accounts
+    UserAccounts::init();
+    // listen for new commands from CLI
+    let mut stdin = io::BufReader::new(io::stdin()).lines();
+
+    // connect the matrix bot with the qaul-cli
+    relay_bot::connect();
```
MathJud: please handle the returned result.

MathJud (commented):
congrats to this first successful integration of the matrix-sdk crate into qaul!
I added a few comments on how to integrate it further into the qaul system.

This will include the configuration system and the use of the RPC messaging system with libqaul:

* check libqaul events using the RPC system
* save the matrix stuff into the same directory as libqaul does, to keep everything in one place.

rust/clients/bridge/src/relay_bot.rs
```diff
@@ -0,0 +1,120 @@
...
+                let content = AnyMessageEventContent::RoomMessage(MessageEventContent::text_plain(
+                    "I am a message sent from qaul network\n",
+                ));
+
+                // send our message to the matrix room we found the "!qaul command
+                room.send(content, None).await.unwrap();
+            }
+
+            if msg_body.contains("!users-list") {
+                // Here I need to trigger the RPC call from users.rs file.
```
Harshil-Jani: @MathJud This is the place where I am interested in calling the rpc function from `users.rs`. 
Use case is that. In matrix room, a user sends a message `!users-list`. This should return the data which we will format with better strings. 

If we are successful to do just this, Next we can call the rpc function of `feed.rs` and that should allow us to send any messages which we write in qaul world to be sent over to the matrix room. And this should be enough to complete our first milestone.

MathJud: There are several ways to do that.

1) The bridge client queries regularly libqaul via RPC for a list of users, e.g. all 10 seconds. This list is stored in a way to be accessed by different threads and functions (see the state storage object https://crates.io/crates/state )

2) Create an asynchronous event system, where you pause your function until the next user list arrived.

As we anyway will need the user information, and also will need to query several other information regularly I would go with solution 1. It is also much simpler to implement.

MathJud: Which users do you want to deliver here?

a) all known users, independent if they are online or not (this is an ever growing list of users)
b) the users that are online at this time.

Harshil-Jani: > Which users do you want to deliver here?
a) all known users, independent if they are online or not (this is an ever growing list of users)
b) the users that are online at this time.

It's (a)

PR #590, opened 2023-08-23
Upgrade to libp2p 0.52

MathJud:
This PR upgrades libqaul to libp2p 0.52.
It also upgrades all rust crates to the current versions.


MathJud:
fixed by #604

PR #602, opened 2023-11-21
fix: improve group creation flow on mobile

brenodt:
## Description
Upon reviewing, I was able to reproduce the issue on Android - on Mac it wasn't reproducible, since Desktop has a different "Chat opening" logic (it's nested in the tab, rather than being a new screen).

I was able to improve the chat creation flow by changing how the navigation is being handled, and returning the newly created room from the "Create Group" dialog.

@MathJud could you please test on Android also to see if I didn't forget anything?

#### Android 10, Moto G7 - bug fixed

https://github.com/qaul/qaul.net/assets/54450520/39221dd4-5540-42fc-a20b-c7bfbac91981

#### MacOS Ventura 13.3.1 (a) - no regression

https://github.com/qaul/qaul.net/assets/54450520/3add08fb-c36a-4c7e-85d1-6cee3a1b2fc6


MathJud:
works great!

Issue #589, opened 2023-08-16
2.0.0-beta.16 and 2.0.0-beta.17 crashes on Android when initializing Bluetooth


link2xt:
I installed https://github.com/qaul/qaul.net/releases/download/v2.0.0-beta.16/qaul-2.0.0-beta.16.apk from the website.

The application keeps crashing before showing any UI.

adb logcat output:
```
08-16 16:32:10.452 25363 28847 I libqaul::connections:..: BLE module start result received
08-16 16:32:10.452 25363 28847 I libqaul::connections:..: BLE Module successfully started
08-16 16:32:10.456  3489 13364 E Parcel  : Attempt to read from protected data in Parcel 0x7a6906ca68
08-16 16:32:10.459 25363 28874 E AndroidRuntime: FATAL EXCEPTION: pool-5-thread-1
08-16 16:32:10.459 25363 28874 E AndroidRuntime: Process: net.qaul.qaul_app, PID: 25363
08-16 16:32:10.459 25363 28874 E AndroidRuntime: java.lang.NullPointerException: Attempt to invoke interface method 'android.os.IBinder android.os.IInterface.asBinder()' on a null object reference
08-16 16:32:10.459 25363 28874 E AndroidRuntime:        at android.os.Parcel.createException(Parcel.java:2077)
08-16 16:32:10.459 25363 28874 E AndroidRuntime:        at android.os.Parcel.readException(Parcel.java:2039)
08-16 16:32:10.459 25363 28874 E AndroidRuntime:        at android.os.Parcel.readException(Parcel.java:1987)
08-16 16:32:10.459 25363 28874 E AndroidRuntime:        at android.bluetooth.IBluetoothGatt$Stub$Proxy.startAdvertisingSet(IBluetoothGatt.java:1526)
08-16 16:32:10.459 25363 28874 E AndroidRuntime:        at android.bluetooth.le.BluetoothLeAdvertiser.startAdvertisingSet(BluetoothLeAdvertiser.java:434)
08-16 16:32:10.459 25363 28874 E AndroidRuntime:        at android.bluetooth.le.BluetoothLeAdvertiser.startAdvertisingSet(BluetoothLeAdvertiser.java:306)
08-16 16:32:10.459 25363 28874 E AndroidRuntime:        at android.bluetooth.le.BluetoothLeAdvertiser.startAdvertising(BluetoothLeAdvertiser.java:158)
[...]

PR #616, opened 2024-03-25
Do not crash if bluetoothLeAdvertiser.startAdvertising fails

link2xt:
On my phone it throws NullPointerException.
This workaround prevents crash on my phone
when bluetooth is enabled.
This will not fix bluetooth support,
but at least the app is usable over Wi-Fi.

Related issue describing the crash
with stacktraces: <https://github.com/qaul/qaul.net/issues/589>


link2xt:

MathJud:
Thanks for this fix!

PR #626, opened 2024-05-03
qaul 2.0.0 Release Candidate 1 preparation

MathJud:
This PR prepares all changes for the qaul 2.0.0 release candidate 1 release

- It only updates from qaul 2.0.0 beta 18
  - as we removed the old data base
- Added various information and helper tools for users who try to update from older versions.
- The TCP stack was removed. Using QUIC as the new interconnection protocol.
  - it is secure by default
  - makes our network smaller
  - has a much better success rate in hole punching
  - we can get rid of a lot of listening interfaces
- The configuration is automatically upgraded to this new version.


brenodt (commented):
LGTM 👍

PR #632, opened 2024-06-07
Fix: network view flicker

brenodt:
## Description
The cause of this behavior resided in the `qaul_rpc` package, which exposes a `StateNotifier<List<User>>`.

On every second, we refresh the users list, so that the UI can be notified in case of new users or state changes for existing users.

This was causing the UI to re-render, since the list of users changed on every second (i.e. the object changed, not the actual list of users).

A state change notifies the UI to trigger a re-render, which is perceived as a flicker.

This is solved by not updating the notifier's state when it and the new list of users are deeply equal.


MathJud:
Thanks a lot! It works great!
This PR was merged into the `libp2p53` branch and merged there :)
I'm therefor closing the PR here.

PR #666, opened 2025-03-06
[#634] Send photos and files without mandatory text message

### LINKED ISSUE #634: Send photos and files without mandatory text message
At the moment the interface does not allow sending a file or photo, without entering a text message.
This should not be the case.

brenodt:
## Description
Closes #634 .


MathJud:
Thanks for this fix!
Would it be possible to remove the blue space under the image, when there is no text, and only display the image then?

MathJud:
Looks great! thanks!

PR #671, opened 2025-07-11
Implement the password support

mohitk108:
Part of https://github.com/qaul/qaul.net/issues/531

Note:
Completed the core logic of adding password support, which includes updating the protobuf definitions. Also, I've been using unwraps in some places, and would add proper error handling at the completion of this PR.
I'm using `qaul_id: PeerId` instead of QaulId, that would be updated

Update:
The commit history is now sane.
The protobuf definitions have been revised based on our IRC chat.

Issue #531, opened 2023-02-10
libqaul <-> UI protobuf RPC authentication & encryption


MathJud:
Create an authentication layer to have authenticated & encrypted communication between the UI and libqaul.

The communication between the GUI and libqaul is done via Protobuf [RPC](https://github.com/qaul/qaul.net/tree/main/rust/libqaul/src/rpc) messages.

The current system shall be extended to allow multiple sessions and multiple users. It shall have a user authentitcation and encrypted communication sessions.

* The authentication should be done via user password or local session key.
* The password shall be optional for the default user.
* The sessions should be transport encrypted via Noise protocol.
* Multiple sessions shall be possible per user.
* Multiple users shall be supported per node.

sukhman-sukh:
Hey @MathJud ,
I am Sukhman Singh.
I would like to work on this issue if it is not solved yet.
Please assign this issue to me .
Thank you.

MathJud:
@sukhman-sukh  Thanks a lot for your interest in this issue!
This is a complex one, as it needs to be solved on the rust side (in libqaul & in the rust CLI clients) and on the flutter side (UI) in parallel.
I suggest to first create a written concept in the discussion in order to have a good plan on how to create and implement a solution.

PR #675, opened 2025-09-26
Add token generation and use it for session's persistence

mohitk108:
Part of https://github.com/qaul/qaul.net/issues/531
Built on top of https://github.com/qaul/qaul.net/pull/671.


brenodt:
Merged via #771.

PR #707, opened 2026-01-13
[#000] Migrate from Hive to SharedPreferences

GustavoDcs:
## Description

This PR removes the unmaintained `Hive` package and migrates user preference storage to the modern `SharedPreferencesWithCache`

## Changes

- Removed packages: `hive`, `hive_flutter`, `hive_generator`
- Migrated `UserPrefsHelper` to use `SharedPreferencesWithCache` (`SharedPreferences` is also being [deprecated](https://pub.dev/packages/shared_preferences))
- All setters now trigger notifications to maintain UI reactivity
- Unit test to cover `UserPrefsHelper`

## Additional Notes
Users will lose their saved preferences related to (language, theme, notifications) on first launch after this update. 

All preferences will reset to system defaults.

Also, @brenodt let me know if there is any file that I shouldn't be including in this PR.

## Evidence

https://github.com/user-attachments/assets/85308e18-0ed8-4de4-8ec0-6f9827b5c778

### DIFF
--- a/qaul_ui/lib/helpers/user_prefs_helper.dart
+++ b/qaul_ui/lib/helpers/user_prefs_helper.dart
@@ -1,19 +1,49 @@
+import 'dart:async';
+
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 import 'package:shared_preferences/shared_preferences.dart';
 
 class UserPrefsHelper {
-  UserPrefsHelper._internal(this._prefs);
+  UserPrefsHelper._internal(this._prefs) {
+    _localeNotifier = ValueNotifier(_loadLocaleFromPrefs());
+    _publicTabNotificationsNotifier = _createBoolNotifier(_publicNTFYKey, defaultValue: true);
+    _chatNotificationsNotifier = _createBoolNotifier(_chatNTFYKey, defaultValue: true);
+    _verifiedUsersOnlyNotifier = _createBoolNotifier(_verifyNTFKey, defaultValue: false);
+  }
+  
+  ValueNotifier<bool> _createBoolNotifier(String key, {required bool defaultValue}) {
+    return ValueNotifier(_prefs.getBool(key) ?? defaultValue);
+  }
   
   static UserPrefsHelper? _instance;
+  static final Completer<void> _readyCompleter = Completer();
+  
   final SharedPreferencesWithCache _prefs;
-  final ValueNotifier<int> _changeNotifier = ValueNotifier(0);
+  
+  late final ValueNotifier<Locale?> _localeNotifier;
+  late final ValueNotifier<bool> _publicTabNotificationsNotifier;
+  late final ValueNotifier<bool> _chatNotificationsNotifier;
+  late final ValueNotifier<bool> _verifiedUsersOnlyNotifier;
+  
+  static Future<void> get ready => _readyCompleter.future;
   
   static Future<void> initialize() async {
+    if (_instance != null) return;
+    
     final prefs = await SharedPreferencesWithCache.create(
       cacheOptions: const SharedPreferencesWithCacheOptions(),
     );
     _instance = UserPrefsHelper._internal(prefs);
+    
+    if (!_readyCompleter.isCompleted) {
+      _readyCompleter.complete();
+    }
+  }
+  
+  @visibleForTesting
+  static void resetForTesting() {
+    _instance = null;
   }
   
   factory UserPrefsHelper() {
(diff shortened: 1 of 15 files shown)


qaul_ui/lib/qaul_app.dart
brenodt: Maybe we drop saving the theme on the user preferences helper to relying solely on the adaptive theme?
https://pub.dev/packages/adaptive_theme#get-thememode-at-app-start

qaul_ui/lib/widgets/language_select_dropdown.dart
```diff
@@ -33,8 +33,8 @@ class _LanguageDropdown extends StatelessWidget {
     final items = <Locale?>[null, ...AppLocalizations.supportedLocales];
 
     return ValueListenableBuilder(
-      valueListenable: Hive.box(UserPrefsHelper.hiveBoxName).listenable(),
-      builder: (context, box, _) {
+      valueListenable: UserPrefsHelper().listenable,
```
brenodt: Perhaps we should avoid relying on a single "listenable" prop to trigger all re-renders and, instead, expose one "notifier" per field of interest (e.g. `localeNotifier` in this case).
This also helps us avoid fetching the value directly (line 39) as we can use the value passed as one of the builder props (line 37) which are now ignored (`_, __`)

qaul_ui/lib/helpers/user_prefs_helper.dart
```diff
@@ -28,31 +47,45 @@ class UserPrefsHelper {
...
   }
 
   set defaultTheme(ThemeMode theme) {
-    _prefsBox?.put(_defaultThemeKey, ThemeMode.values.indexOf(theme));
+    _prefs.setInt(_defaultThemeKey, ThemeMode.values.indexOf(theme));
+    _notifyChange();
   }
 
-  bool get publicTabNotificationsEnabled => _prefsBox?.get(_publicNTFYKey) ?? true;
+  bool get publicTabNotificationsEnabled => _prefs.getBool(_publicNTFYKey) ?? true;
```
brenodt: Please double-check that we don't need change notifiers for these three booleans (or if we have doubled state variables, private on the settings page).

qaul_ui/test/user_prefs_helper_test.dart
```diff
@@ -0,0 +1,73 @@
...
+
+void main() {
+  TestWidgetsFlutterBinding.ensureInitialized();
+
+  setUp(() async {
+    SharedPreferencesAsyncPlatform.instance = InMemorySharedPreferencesAsync.empty();
+    await UserPrefsHelper.initialize();
+  });
+
+  group('Hive to SharedPreferences migration', () {
```
brenodt: Let's rename this group, we don't need to refer to the dependency we're dropping. It's a valid test group regardless.

brenodt (changes requested):
Minor comments, overall it's looking great! Kudos

GustavoDcs:

qaul_ui/lib/qaul_app.dart
GustavoDcs: You're right, I removed it!

qaul_ui/lib/helpers/user_prefs_helper.dart
GustavoDcs: I double checked it and there is no duplicated state. The ValueNotifiers live in UserPrefsHelper as the single source of truth, getters read from them, setters persist to disk AND update them. 

The settings screen doesn't have private state; it reads directly from UserPrefsHelper and uses setState() for local rebuilds.

qaul_ui/lib/widgets/language_select_dropdown.dart
GustavoDcs: Great suggestion, already done! I've implemented granular notifiers (one per preference) in UserPrefsHelper.

This eliminates the need for a single "trigger all" listenable and ensures widgets only rebuild when their specific preference changes.

qaul_ui/test/user_prefs_helper_test.dart
GustavoDcs: Done!

GustavoDcs:
Moved to draft because I found a bug involving the Linux application.

qaul_ui/lib/helpers/user_prefs_helper.dart
```diff
@@ -1,58 +1,155 @@
...
+  
+  final _prefs = SharedPreferencesAsync();
+  
+  final _localeNotifier = ValueNotifier<Locale?>(null);
+  final _themeModeNotifier = ValueNotifier<ThemeMode>(ThemeMode.system);
+  final _publicTabNotificationsNotifier = ValueNotifier<bool>(true);
+  final _chatNotificationsNotifier = ValueNotifier<bool>(true);
+  final _verifiedUsersOnlyNotifier = ValueNotifier<bool>(false);
+  
+  static Future<void> get ready => _readyCompleter.future;
```
brenodt: Since this isn't used, we can simplify this object's implementation by removing it and the Completer that drives it.

MathJud:
fixes #744

PR #714, opened 2026-01-22
 [Version] Remove unsupported project dependencies and recreate them internally

GustavoDcs:
## Description
This PR removes the external `version` package dependency and internalizes its implementation directly into the repository.

The `version` package code was fully replicated into the codebase, preserving its original behavior. This allows us to eliminate the external dependency while maintaining full compatibility with existing usage.

### Dependency Removal: `version`
- Removed `version` from `pubspec.yaml`

### Internalization of `version` Package
- Replicated the full source code of the original `version` package into the repository

### DIFF
--- a/qaul_ui/packages/utils/lib/src/version.dart
+++ b/qaul_ui/packages/utils/lib/src/version.dart
@@ -49,7 +49,7 @@ class Version implements Comparable<Version> {
             "preRelease segments must only contain [0-9A-Za-z-]");
       }
     }
-    if (this.build.isNotEmpty && !_buildRegex.hasMatch(this.build)) {
+    if (build.isNotEmpty && !_buildRegex.hasMatch(build)) {
       throw FormatException("build must only contain [0-9A-Za-z-.]");
     }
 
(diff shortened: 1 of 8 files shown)


qaul_ui/packages/utils/lib/src/version.dart
brenodt, line 1: Let's cleanup this file, as there is a lot here that we don't make use of, and we can always re-introduce if needed.
For all public methods/fields, if they aren't used internally, we can remove them.

qaul_ui/packages/utils/lib/src/version.dart
```diff
@@ -0,0 +1,261 @@
+// Copyright (c) 2021, Matthew Barbour. All rights reserved. Use of this source code
+// is governed by a BSD-style license that can be found in the LICENSE file.
+
+/// Provides version objects to enforce conformance to the Semantic Versioning 2.0 spec. The spec can be read at http://semver.org/
+library;
```
brenodt:
```suggestion
```

brenodt (changes requested):
A bit of refactoring is in order before merging this one. We can reduce the `Version` API a bit, as our use-case is quite constrained and doesn't require much of it.

qaul_ui/packages/utils/lib/src/version.dart
GustavoDcs: Done!

qaul_ui/packages/utils/lib/src/version.dart
GustavoDcs: Done!

MathJud:
fixes #744

PR #715, opened 2026-01-22
 [Mailto] Remove unsupported project dependencies and recreate them internally

GustavoDcs:
## Description
This PR removes the external `mailto` package dependency and internalizes its implementation directly into the repository.

The `mailto` package code was fully replicated into the codebase, preserving its original behavior. This allows us to eliminate the external dependency while maintaining full compatibility with existing usage.

### Dependency Removal: `mailto`
- Removed `mailto` from `pubspec.yaml`

### Internalization of `mailto` Package
- Replicated the full source code of the original `version` package into the repository


brenodt (changes requested):
As we don't use anything but the string encoding, I believe we can do without this library entirely.

How about we refactor the code in `qaul_ui/lib/coordinators/email_logging_coordinator/email_logging_coordinator.dart` to use a simple `encodeMailto(Uri uri)` instead?

You can make it part of `utils` or keep it on the main app code, I don't mind either way.

Some sources that can be of help:
- https://en.wikipedia.org/wiki/Mailto
- https://www.rfc-editor.org/rfc/rfc6068.html
- https://github.com/dart-lang/sdk/issues/43838#issuecomment-827060288

I suggest we also add a few unit tests, as this has very specific intended behavior and it shouldn't take long to validate the intended behavior.

MathJud:
fixes #744

PR #718, opened 2026-01-23
[Open Simplex2] Remove unsupported project dependencies and recreate them internally

GustavoDcs:
## Description
This PR removes the external `open_simplex2` package dependency and internalizes its implementation directly into the repository.

The `open_simplex2` package code was fully replicated into the codebase, preserving its original behavior. This allows us to eliminate the external dependency while maintaining full compatibility with existing usage.

### Dependency Removal: `open_simplex2`
- Removed `open_simplex2` from `pubspec.yaml`

### Internalization of `open_simplex2` Package
- Replicated the full source code of the original `open_simplex2` package into the repository


MathJud:
fixes #744

PR #719, opened 2026-01-28
Upgrade libp2p v0.56.0

dastansam:
- deprecates async-std and uses tokio instead
- other changes related to libp2p 56.0
- adds e2e testing script
- adds research docs


MathJud:
fixes #730

MathJud:
fixes #729

PR #722, opened 2026-02-05
Fix chat UI not updating when opening conversation (Riverpod 3)

GustavoDcs:
## Description
This PR fixes the chat not showing messages when opening a conversation after the Riverpod 3 migration.

In Riverpod 3, all providers use `==` to decide whether to notify listeners. With `ChatRoom` using only `idBase58` in its Equatable `props`, an updated room (with new messages) was considered equal to the previous one, so the provider did not notify and the chat UI did not rebuild. This change aligns `ChatRoom` equality and room lookup with that behaviour, and fixes the comparison and update logic in the chat translator so the open room receives and displays messages correctly.

### ChatRoom equality (Riverpod 3 compatibility)
- In `chat_room.dart`, extended `props` to `[idBase58, lastMessageIndex]` so state changes when new messages arrive and the provider notifies and the UI rebuilds
- Kept only `lastMessageIndex` (not the full `messages` list) in `props` to avoid heavy comparisons in long chats

### Chat translator
- In `chat_translator.dart`, compare `conversationId` and `groupId` with `ListEquality<int>().equals()` instead of `.equals()` so the same conversation is recognised and the correct room is updated
- When the open room is not yet in `chatRoomsProvider` but the response matches `currentOpenChatRoom`, merge the response, add the room to the list, and set `currentOpenChatRoom` so the open chat shows messages

### Room identification by `idBase58`
- In `chat_room_list_notifier.dart`, `contains` and `update` now identify rooms by `idBase58` so the right room is found and replaced even when the room instance changes
- In `group_settings.dart`, room lookup uses `firstWhere((r) => r.idBase58 == room.idBase58)` so the correct room is used from the list

## Evidence
https://github.com/user-attachments/assets/e039fbc1-6402-466b-be5a-033f245e3107

https://github.com/user-attachments/assets/ff16ffa2-401c-44fc-99c1-c54fc233f3ae

### DIFF
--- a/qaul_ui/lib/screens/home/tabs/chat/widgets/group_settings.dart
+++ b/qaul_ui/lib/screens/home/tabs/chat/widgets/group_settings.dart
@@ -10,7 +10,7 @@ class _GroupSettingsPage extends HookConsumerWidget {
   @override
   Widget build(BuildContext context, WidgetRef ref) {
     final defaultUser = ref.watch(defaultUserProvider)!;
-    final group = ref.watch(chatRoomsProvider).firstWhere((r) => r == room);
+    final group = ref.watch(chatRoomsProvider).firstWhere((r) => r.idBase58 == room.idBase58);
 
     final nameCtrl = useTextEditingController(text: room.name);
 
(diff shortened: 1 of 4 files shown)


brenodt (approved):
LGTM

MathJud:
fixes #744

Issue #541, opened 2023-03-22
create qauld-ctl to control qauld


MathJud:
The qaul daemon [qauld](https://github.com/qaul/qaul.net/tree/main/rust/clients/qauld) runs headless in the background. Create a second binary `qauld-ctl` To control it. For this you also need to setup a communication channel between `qauld` and `qauld-ctl` over which you can control qauld via the protobuf RPC.

- [x] Create a local communication channel for qauld which is only locally accessible (e.g. a unix socket)
- [x] Create a new `qauld-ctl` CLI binary that connects to `qauld`
- [x] Send protobuf messages from `qauld-ctl` to `qauld` and display the results in `qauld-ctl`
  - For how to use the protobuf RPC to control libqaul, check [qaul-cli](https://github.com/qaul/qaul.net/tree/main/rust/clients/cli)

Jadhav-Dhanashri:
Hello maintainers,
I'm new to freifunk.net open source contributions excited to get started. I would like to work on this issue , & I'll follow the contribution guidelines carefully. Could you please Assign it to me if it's available?
Thank you!

Lilit0x:
@MathJud 

Basically, we need a way to control running `qauld` instances programmatically. The current qaul-cli embeds libqaul directly and communicates via in-process channels, which is okay for a standalone client, but cannot control an already-running daemon.

The goal is two-fold:
1. Add a control interface to qauld (Unix socket)
2. Create `qauld-ctl` binary that connects to it

Primary use case: the e2e testing framework. Secondary: giving users a proper CLI to manage their daemon.

## Where Should The Scoket Be Located?

We can run multiple `qauld` instances on the same machine, and each needs its own socket. The earlier [GSoC attempt](#543) hardcoded `/tmp/qauld.sock`, which obviously breaks with multiple instances.

So, you suggested `/var/run/qauld/qauld.sock` with user-specific subdirectories, but that has permission issues for non-root users.

Here is what I suggest: we put the socket in the `pwd`, same place `qauld` already stores its logs, config file and the rest. Each instance runs from a different directory, so isolation is automatic.

```
/path/to/instance-1/
├── config.yaml
├── qauld.sock    # control socket
├── qauld.pid     # for process management
└── logs/

/path/to/instance-2/
├── config.yaml
├── qauld.sock    ← another different seperate socket
├── qauld.pid
└── logs/
```

qauld-ctl finds the socket by:
1. Explicit `--socket /path/to/qauld.sock`
[...]

PR #747, opened 2026-02-12
Add qauld-ctl CLI and Unix socket control interface for qauld (#541)

Lilit0x:
## Description
  This PR adds the ability to control a running qauld daemon from external processes. Currently, the only CLI client (qaul-cli) embeds libqaul directly and communicates via in-process channels, which means it can't control an already-running daemon. This work introduces a Unix socket control interface on qauld and a new qauld-ctl binary that connects to it, primarily to support the e2e testing framework.

  The socket server in qauld listens on `./qauld.sock` in the working directory (the same folder where config and database live), which naturally supports multiple instances since each runs from a different directory. It uses length-prefixed protobuf framing over the socket, spawns a task per client connection, and has a central RPC poller that routes responses back to the correct client usin `request_id` tracking.

  The qauld-ctl binary currently supports single-shot mode — connect, send a command, print the response, and exit. It discovers the socket through `--socket`, `--dir`, the current directory, or the `QAULD_SOCKET` environment variable.

  To make response routing work, all RPC handler modules in libqaul were updated to accept and forward `request_id` through to their responses. Previously, responses were sent back with an empty `request_id`, making it impossible to match them to the requesting client.

TODO:
- [x] More CLI commands (user accounts, users, chat, feed, etc.)
- [ ] `--json` output flag
- [x] PID file management (qauld.pid): no longer needed, I think `qaukd.sock` gets the job done enough.
- [ ]  Interactive CLI mode (qauld-ctl -i)

#541


Lilit0x:

rust/clients/qauld/src/main.rs
```diff
@@ -75,8 +77,9 @@ async fn main() {
         libqaul::node::user_accounts::UserAccounts::create(user_name.clone(), None);
     }
 
-    // loop
-    loop {
-        thread::sleep(Duration::from_millis(10));
+    // since we're now starting a socket server, the main thread now has work to do,
```
MathJud, line 80: Please create a description of what it does, not what it does not anymore :)
e.g.: start listening on unix socket

MathJud (commented):
Looks great!
Thanks that you implemented the `request_id` this was a long waiting todo.

PR #748, opened 2026-02-13
BLE GATT messaging protocol

MathJud:
New BLE GATT messaging protocol

This protocol creates a two Byte header for all GATT messages.
The procotol can:

- cut qaul messages in chunks and reassemble the message
- detect missing chunks and reorder them
- send acknowledgements for received and failed packages


MathJud:
Work was continued in #908

PR #745, opened 2026-02-08
Users pagination support

brenodt:
## Description
This PR aims to add pagination support on libaqul to the two methods for retrieving users in the Users rpc module:
- `UserRequest`
- `UserOnlineRequest`

_Note_: the generated files can be largely ignored for the purposes of this PR review:
- qaul_ui/packages/qaul_rpc/lib/src/generated/router/users.pb.dart
- qaul_ui/packages/qaul_rpc/lib/src/generated/router/users.pbenum.dart
- qaul_ui/packages/qaul_rpc/lib/src/generated/router/users.pbjson.dart
- rust/libqaul/src/router/qaul.rpc.users.rs
- rust/libqaul/src/rpc/protobuf_generated/rust/qaul.rpc.users.rs

### What's changed

`rust/libqaul/src/router/users.proto`:
- define new `PaginationMetadata` message

`rust/libqaul/src/router/users.rs`:
- unify the implementation of `UserRequest` and `UserOnlineRequest` by:
  - reducing duplicated code;
  - defining a private method to retrieve state-dependent fields of the implementation;
  - defining a private, stateless function to enable unit testing.
- unit test the core user list building logic;
- add a backwards-compatible implementation of pagination to the two methods

PR #749, opened 2026-02-14
Implement User Pagination

GustavoDcs:
## Description

This PR adds pagination to the user list in the Users tab. Instead of loading all users at once, the app now requests users in pages ([thanks to this change](https://github.com/qaul/qaul.net/pull/745)). When the user scrolls near the bottom of the list, the next page is loaded and appended. Pull-to-refresh resets to the first page. This reduces initial load and memory use when many users are present.

### Users tab UI (`users_tab.dart`)
- Initial load requests the first page (`offset: 0`, `limit: _pageSize` with `_pageSize = 50`) in a post-frame callback.
- Scroll listener triggers `_loadMoreUsers()` when the user reaches 80% of the scroll extent; it requests the next page using the current offset and the same limit.
- Pull-to-refresh resets offset and `hasMore`, then requests the first page again.
- `ref.listen(usersPaginationStateProvider)` updates `_hasMore` when the backend reports no more pages.
- `_updatePaginationState()` keeps `_currentOffset` and `_hasMore` in sync with the last pagination response (including after `getUsers` returns).
- List shows a loading indicator at the bottom while a page is loading and hides it when there are no more pages.

### DIFF
--- a/qaul_ui/lib/screens/home/tabs/users_tab.dart
+++ b/qaul_ui/lib/screens/home/tabs/users_tab.dart
@@ -136,7 +136,6 @@ class _UsersState extends _BaseTabState<_Users> {
             }),
           ),
         ),
-      ),
     );
   }
 }
--- a/qaul_ui/lib/screens/home/tabs/users_tab.dart
+++ b/qaul_ui/lib/screens/home/tabs/users_tab.dart
@@ -83,34 +83,23 @@ class _UsersState extends _BaseTabState<_Users> {
 
     final l10n = AppLocalizations.of(context)!;
     return Scaffold(
-      body: RefreshIndicator(
-        onRefresh: () async => await _refreshUsers(),
-        child: SearchUserDecorator(builder: (_, users) {
-            return EmptyStateTextDecorator(
-              l10n.emptyUsersList,
-              isEmpty: users.isEmpty,
-              child: ListView.separated(
-                controller: _scrollController,
-                physics: const AlwaysScrollableScrollPhysics(),
-                itemCount: users.length + (_hasMore.value ? 1 : 0),
-                separatorBuilder: (_, _) => const Divider(height: 12.0),
-                itemBuilder: (_, i) {
-                  if (i == users.length) {
-                    return ValueListenableBuilder<bool>(
-                      valueListenable: _isLoadingMore,
-                      builder: (context, isLoading, _) {
-                        if (isLoading) {
-                          return const Padding(
-                            padding: EdgeInsets.all(16.0),
-                            child: Center(child: LoadingIndicator()),
-                          );
-                        }
-                        return const SizedBox.shrink();
-                      },
-                    );
-                  }
-
-                  final user = users[i];
+      body: ListenableBuilder(
+        listenable: _isLoadingMore,
+        builder: (context, _) => LoadingDecorator(
+          isLoading: _isLoadingMore.value,
+          child: RefreshIndicator(
+            onRefresh: () async => await _refreshUsers(),
+            child: SearchUserDecorator(builder: (_, users) {
+              return EmptyStateTextDecorator(
+                l10n.emptyUsersList,
+                isEmpty: users.isEmpty,
+                child: ListView.separated(
+                  controller: _scrollController,
+                  physics: const AlwaysScrollableScrollPhysics(),
+                  itemCount: users.length,
+                  separatorBuilder: (_, _) => const Divider(height: 12.0),
+                  itemBuilder: (_, i) {
+                    final user = users[i];
                   var theme = Theme.of(context).textTheme;
                   var hasConnections =
                       user.availableTypes != null && user.availableTypes!.isNotEmpty;
@@ -145,10 +134,12 @@ class _UsersState extends _BaseTabState<_Users> {
                       tapRoutesToDetailsScreen: true,
                     ),
                   );
-                },
-              ),
-            );
-        }),
+                  },
+                ),
+              );
+            }),
+          ),
+        ),
       ),
     );
   }
--- a/qaul_ui/lib/screens/home/tabs/users_tab.dart
+++ b/qaul_ui/lib/screens/home/tabs/users_tab.dart
@@ -8,31 +8,117 @@ class _Users extends BaseTab {
 }
 
 class _UsersState extends _BaseTabState<_Users> {
+  static const _pageSize = 50;
+  late final ScrollController _scrollController;
+  final _isLoadingMore = ValueNotifier<bool>(false);
+  final _hasMore = ValueNotifier<bool>(true);
+  int _currentOffset = 0;
+
+  @override
+  void initState() {
+    super.initState();
+    _scrollController = ScrollController();
+    _scrollController.addListener(_onScroll);
+    WidgetsBinding.instance.addPostFrameCallback((_) {
+      _refreshUsers();
+    });
+  }
+
+  @override
+  void dispose() {
+    _scrollController.removeListener(_onScroll);
+    _scrollController.dispose();
+    _isLoadingMore.dispose();
+    _hasMore.dispose();
+    super.dispose();
+  }
+
+  void _onScroll() {
+    if (_scrollController.position.pixels >=
+        _scrollController.position.maxScrollExtent * 0.8) {
+      _loadMoreUsers();
+    }
+  }
+
+  void _updatePaginationState() {
+    final paginationState = ref.read(usersPaginationStateProvider);
+    if (paginationState != null) {
+      _hasMore.value = paginationState.hasMore;
+      _currentOffset = paginationState.offset + paginationState.limit;
+      return;
+    }
+    _currentOffset += _pageSize;
+  }
+
+  Future<void> _loadMoreUsers() async {
+    if (_isLoadingMore.value || !_hasMore.value) return;
+
+    _isLoadingMore.value = true;
+    try {
+      final worker = ref.read(qaulWorkerProvider);
+      await worker.getUsers(offset: _currentOffset, limit: _pageSize);
+      _updatePaginationState();
+    } finally {
+      _isLoadingMore.value = false;
+    }
+  }
+
+  Future<void> _refreshUsers() async {
+    _currentOffset = 0;
+    _hasMore.value = true;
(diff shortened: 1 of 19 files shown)


qaul_ui/packages/qaul_rpc/lib/src/rpc_translators/users_translator.dart
```diff
@@ -78,12 +91,31 @@ class UsersTranslator extends RpcModuleTranslator {
...
+      if (!isFirstPage) {
+        usersNotifier.appendMany(paginatedUsers.users);
+      }
+      ref.read(usersPaginationStateProvider.notifier).setPagination(
+            paginatedUsers.pagination,
+          );
+      return;
+    }
+    
     if (res.data is List<User>) {
```
brenodt: This branch is no longer possible as there are no cases in this translator's `decodeMessageBytes` where we return a `List<User>`.

We can safely remove it.

qaul_ui/packages/qaul_rpc/lib/src/rpc_translators/users_translator.dart
```diff
@@ -78,12 +91,31 @@ class UsersTranslator extends RpcModuleTranslator {
...
     if (res.module != type) return;
+    
+    if (res.data is PaginatedUsers) {
+      final paginatedUsers = res.data as PaginatedUsers;
+      final usersNotifier = ref.read(usersProvider.notifier);
+      final isFirstPage = paginatedUsers.pagination?.offset == 0;
+      if (isFirstPage) {
+        usersNotifier.replaceAll(paginatedUsers.users);
+      }
+      if (!isFirstPage) {
```
brenodt: Maybe this would be a good use-case for an `else`? A bit more readable imo

qaul_ui/packages/qaul_rpc/lib/src/models/user.dart
```diff
@@ -12,6 +12,20 @@ final usersProvider = NotifierProvider<UserListNotifier, List<User>>(
   UserListNotifier.new,
 );
 
+class UsersPaginationStateNotifier extends Notifier<UsersPaginationState?> {
```
brenodt: I'd argue that we shouldn't add a new `provider` here: if we do, we're accepting that for all features that will support pagination, we will add `paginationStateNotifiers` to accommodate them.

Perhaps we need an abstraction to merge the state of the pagination metadata _and_ the content being paginated.

It will require a refactor on all places that access the content, but it should be as simple as changing the provider name and accessing a field (instead of the raw `usersProvider` value).

Let me know if you agree.

qaul_ui/packages/qaul_rpc/lib/src/models/user.dart
```diff
@@ -132,26 +146,59 @@ class UserListNotifier extends Notifier<List<User>> {
...
-                u.status == ConnectionStatus.offline ? usr.status : u.status,
-            keyBase58: u.keyBase58 ?? usr.keyBase58,
-            isBlocked: u.isBlocked ?? usr.isBlocked,
-            isVerified: u.isVerified ?? usr.isVerified,
-            availableTypes: u.availableTypes ?? usr.availableTypes,
-          )
-        else
-          usr,
-    ];
+    state = state.map((usr) {
```
brenodt: Could you guide me through your reasoning for the change on `update`?
I'd like to better understand what drove you to it!

qaul_ui/lib/screens/home/tabs/users_tab.dart
```diff
@@ -8,31 +8,117 @@ class _Users extends BaseTab {
...
               child: ListView.separated(
-                controller: ScrollController(),
+                controller: _scrollController,
                 physics: const AlwaysScrollableScrollPhysics(),
-                itemCount: users.length,
+                itemCount: users.length + (_hasMore.value ? 1 : 0),
                 separatorBuilder: (_, _) => const Divider(height: 12.0),
                 itemBuilder: (_, i) {
+                  if (i == users.length) {
+                    return ValueListenableBuilder<bool>(
```
brenodt: What about reusing the `qaul_ui/lib/decorators/loading_decorator.dart` ?

This keeps the UI consistent and makes future updates more centralized.

qaul_ui/lib/screens/home/tabs/users_tab.dart
```diff
@@ -8,31 +8,117 @@ class _Users extends BaseTab {
...
+        _hasMore.value = false;
+      }
+    });
 
     final l10n = AppLocalizations.of(context)!;
     return Scaffold(
       body: CronTaskDecorator(
         schedule: const Duration(milliseconds: 1000),
-        callback: () async => await refreshUsers(),
+        callback: () async {
```
brenodt: This shouldn't be needed anymore: the `CronTaskDecorator` is meant to be used as a polling system, which goes against the pagination approach.

If understand the intent correctly, this implementation would remove the cached user list and reset it once every second, but I don't see why we'd want that.

Let me know if I'm missing something.

brenodt (changes requested):
I see a few topics that I'd like to have a chat about, but overall great effort!

qaul_ui/lib/screens/home/tabs/users_tab.dart
GustavoDcs: I agree with you, that makes sense. My bad, I completely forgot about that detail regarding the CronTaskDecorator and the polling behavior.

Fixed!

qaul_ui/lib/screens/home/tabs/users_tab.dart
GustavoDcs: Good point, I had based this on other pages that were using a `CircularProgressIndicator` for loading, so I followed the same approach here.

Should we create a task to standardize all the loadings in the ui later?

GustavoDcs: Btw, I've changed it.

qaul_ui/packages/qaul_rpc/lib/src/models/user.dart
GustavoDcs: While I was trying to understand the user model, I found it difficult to reason about what this method was doing just by looking at it.

So I tried improving it, I was mainly trying to make the method a bit more readable. Also the single-line else inside the for triggered me a bit visually 🙃, so I tried with a map version that felt clearer to follow at first glance.

If you don't agree or think its an unnecessary change I can revert it, just let me know!

qaul_ui/packages/qaul_rpc/lib/src/rpc_translators/users_translator.dart
GustavoDcs: Agree, I was used to only using `if`s...
Fixed!

qaul_ui/packages/qaul_rpc/lib/src/models/user.dart
GustavoDcs: I tried one approach let me know what you think.

qaul_ui/packages/qaul_rpc/lib/src/models/user.dart
brenodt: All fine! Strictly speaking, the former is a very conventional way of building lists:
https://dart.dev/language/collections#for-element

In theory, it's a bit more performant than a `[].map(...).toList()`: as I interpret it, the functional approach has one extra allocation step. It's certainly negligible for our use-case, but worth considering.

I do agree that the `else` at the end of the former implementation isn't the most readable, though. Would it be better if we early returned the unchanged `usr`? Or would you still think the functional approach is the more readable for this situation?

qaul_ui/packages/qaul_rpc/lib/src/models/user.dart
```diff
@@ -87,30 +111,81 @@ class ConnectionInfo extends Equatable {
...
-  /// If [users] and [state] are deeply equal, will do nothing. As a result, it
-  /// avoids re-rendering UI code that depends on the [List<User>] that this
-  /// notifier exposes.
-  ///
-  /// New users get appended to the list, whilst existing ones get their data
-  /// updated.
-  void updateMany(List<User> users) {
-    if (const ListEquality().equals(state, users)) {
+  void updateMany(List<T> items) {
+    if (const ListEquality().equals(state.data, items)) {
```
brenodt: This method, overall, could be improved (we could do better than a linear check per item on `updatedItems`, post list reallocation).

But, at a minimum, I'd remove this early return:

```suggestion
```

[ListEquality.equals](https://api.flutter.dev/flutter/package-collection_collection/ListEquality/equals.html) will only help in the exact-match case (same items, same order), and even then you've already paid O(n) just to discover you can skip.

In the common case where lists differ, it's pure overhead.

qaul_ui/lib/screens/home/tabs/users_tab.dart
brenodt: Seems to me that we could refactor the code to use the actual `LoadingDecorator` as opposed to rely on this `ValueListenableBuilder`, what do you say?

brenodt (commented):
2 small comments. Non-blocking, but would like to hear your input on them.

qaul_ui/packages/qaul_rpc/lib/src/models/user.dart
brenodt: Great one!

qaul_ui/lib/screens/home/tabs/users_tab.dart
GustavoDcs: Got it, fixed!

GustavoDcs: Done!

qaul_ui/packages/qaul_rpc/lib/src/models/user.dart
GustavoDcs: I tried refactoring even further, separating the methods and returning the for, what you think?

qaul_ui/packages/qaul_rpc/lib/src/models/user.dart
GustavoDcs: See what you think!

brenodt (approved):
LGTM

PR #764, opened 2026-02-19
[Part 2] Navbar Redesign - Refactor

GustavoDcs:
#### Current behaviour:
`_overflowMenuOptions` and `_handleClick` use plain strings for menu/tab-related values even though `TabTyp`e (and similar concepts) already exist.

#### Required change:
Stop using string literals for these cases; use enums instead.
Note: Enums can be turned into strings via the .name property (e.g. Color.blue.name → 'blue'), so there is no need to keep string-based workarounds.

---

### Replace private bar builders with SafeArea + LayoutBuilder and extract QaulNavBar
Current behaviour: build delegates to _buildHorizontalBar and _buildVerticalBar, which rely on MediaQuery for layout/safe area.

#### Required change:

Use a `SafeArea` + `LayoutBuilder` so layout and safe insets are driven by the widget tree instead of manual `MediaQuery` checks.
Extract a `QaulNavBar` widget from the logic currently inside `_buildHorizontalBar` / `_buildVerticalBar`, so the bar is a single, reusable component that can be tested in isolation.

----

### Adjust tests
`QaulNavBarDecorator` tests: Refocus on integration: decorator composes the bar and content, manages global navigation state (e.g. overflow menu → routes), and passes the correct key to the child.

### DIFF
--- a/qaul_ui/lib/decorators/qaul_nav_bar_decorator.dart
+++ b/qaul_ui/lib/decorators/qaul_nav_bar_decorator.dart
@@ -10,6 +10,14 @@ import '../providers/providers.dart';
 import '../screens/home/tabs/tab.dart';
 import '../widgets/widgets.dart';
 
+enum NavBarOverflowOption {
+  settings,
+  about,
+  license,
+  support,
+  oldNetwork,
+  files,
+}
 
 class QaulNavBarDecorator extends StatefulWidget {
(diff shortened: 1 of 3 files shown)


GustavoDcs:
This PR was closed due to it being merged inside the https://github.com/qaul/qaul.net/pull/766

PR #772, opened 2026-02-22
Add a User Profile with mutli-hop DTN routes 

MathJud:
Adds a signable and versionable user profile.
A user profile contains the user name, profile text, the users public key as well as user defined DTN routes.


MathJud:
fixed by #832

PR #816, opened 2026-03-24
Account management

brenodt:


rust/libqaul/src/node/account_management.rs
```diff
@@ -0,0 +1,430 @@
...
+pub struct ExportManifest {
+    pub qaul_version: String,
+    pub sled_version: String,
+    pub export_date: String,
+    pub user_id: String,
+}
+
+/// Serializable subset of the config UserAccount (excludes session_token).
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub struct ExportedAccount {
```
MathJud, line 44: I suggest to also export the users authentication session_token.
The reason behind would be qauld server installations where we deploy that and already have automated clients with existing session tokens that can then further be used.

rust/libqaul/Cargo.toml
```diff
@@ -57,6 +57,8 @@ crc = "3.2"
 fs_extra = "1.3"
 semver = "1.0"
 argon2 = "0.5.3"
+tar = "0.4"
+flate2 = "0.2"
```
MathJud, line 61: pleas upgrade to flate2 1.1
This also defaults to miniz_oxide which is writen in safe rust (right now it compile the C library miniz_sys)

MathJud (commented):
works truly great!

PR #823, opened 2026-03-28
feat(qaul): dtn v2 implementation

dastansam:
- adds dtn v2 with custody routes for a message. maximum 5 diff routes are allowed and max of 10 custodians per route.
- also removes all the unsafe `unwrap()` calls from the codebase


dastansam:
- should also look for routes from common groups between sender and recipient
- garbage collection for expired and undelivered messages

PR #858, opened 2026-04-26
Echo-back feed's send response to enable future-based client implementation

brenodt:


MathJud:
@brenodt 
Please also handle this new feed message in the `qaul-cli` and `qauld-ctl` clients.

At the moment the following error is thrown at `qaul-cli`:

```
ERROR qaul_cli::feed                 > unprocessable RPC feed message
```

PR #860, opened 2026-04-30
Multiple Encrypted Chat Messages during Session Creation

### LINKED ISSUE #846: Create Concept for Multiple Encrypted Chat Messages during Session Creation
At the moment, only one message can be sent by the crypto session initializing party, before the noise RTT has been completed and the symmetric key has been generated.

This leads to bad UX, if the other party is not reachable and messages need to be sent via DTN.

Noise KK allows to send more than one encrypted message during handshake.
Create a concept that allows for multiple messages from the initializing party, initializing the handshake.

dastansam:
closes #846

### DIFF
--- a/docs/proposals/Handshake-Extras-During-Session-Creation.md
+++ b/docs/proposals/Handshake-Extras-During-Session-Creation.md
@@ -1,0 +1,376 @@
+# Multiple Encrypted Chat Messages during Session Creation
+
+## Overview
+
+We should let the initiator send **multiple encrypted payloads on the first
+handshake leg of Noise KK**, queued as additional pre-completion frames under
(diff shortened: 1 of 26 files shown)


MathJud:
when testing it via `qauld-ctl`, I get the following error message in `qauld`:

```txt
 ERROR libqaul::services::crypto    > Can't send further messages after handshake
 ERROR libqaul::services::chat::message > chat message send error Encryption error occurred
 ERROR libqaul::services::crypto        > Can't send further messages after handshake
 ERROR libqaul::services::chat::message > chat message send error Encryption error occurred
 ERROR libqaul::services::crypto        > Can't send further messages after handshake
 ERROR libqaul::services::chat::message > chat message send error Encryption error occurred
```

PR #872, opened 2026-05-15
feat: qauldctl revamp

dastansam:
this is a baseline qauld-ctl revamp that ports most of the cli commands to it.


MathJud:
Sending a feed message with the `qauld-tui` client throws the following error in `qauld`:

```rust
 ERROR libqaul::services::feed      > user account id couldn't be decoded: InvalidMultihash(Error { kind: Io(Kind(UnexpectedEof)) })
```

PR #881, opened 2026-06-04
Feat/implement duplicate name

GustavoDcs:
## Description
Adds duplicate-username metadata to group chat timelines when a joiner’s display name collides with an existing member, following the group-chat redesign.

## Evidence
image


brenodt:
Replaced by #910

PR #888, opened 2026-06-17
Migrate user methods

GustavoDcs:
## Description
Migrates remaining users flows to the future-based RPC path.

### DIFF
--- a/qaul_ui/lib/stores/users_store.dart
+++ b/qaul_ui/lib/stores/users_store.dart
@@ -149,6 +149,8 @@ class UsersSearchStore extends Notifier<UsersSearchState> {
 class UsersStore extends Notifier<List<User>> {
   PaginationState? _pagination;
   PaginationState? get pagination => _pagination;
+  SecurityNumber? _securityNumber;
+  SecurityNumber? get securityNumber => _securityNumber;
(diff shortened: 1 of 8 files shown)


qaul_ui/packages/qaul_rpc/lib/src/rpc_translators/users_translator.dart
brenodt, line 1: Having to comment out the `processResponse` is a code smell. We should aim to move away from this class signature or at least make it opt-in, so that we don't need to keep doing this any further.
Not blocking over this, but please do a refactor on this and the feed module, @GustavoDcs .

qaul_ui/lib/stores/users_store.dart
```diff
@@ -254,6 +256,40 @@ class UsersStore extends Notifier<List<User>> {
     }
   }
 
+  Future<SecurityNumber?> getUserSecurityNumber(User user) async {
+    final securityNumber =
+        await ref.read(qaulWorkerProvider).getUserSecurityNumber(user);
```
brenodt, line 261: Extract ref.read into a variable or a getter to make the code less verbose.

brenodt (approved):

PR #906, opened 2026-07-03
fix(dtn-v2): release hardening — retention cap, quota self-heal, blocked senders, recipient-first check

dastansam:


rust/libqaul/src/services/dtn/mod.rs
```diff
@@ -90,6 +93,26 @@ pub struct DtnStorageStateV2 {
 /// Maximum bytes a single sender can store on this node (10 MB)
 const V2_PER_SENDER_QUOTA: u64 = 10 * 1024 * 1024;
 
+/// Maximum time a V2 custody entry is retained when the sender specified
+/// no expiry (expires_at == 0), counted from local acceptance: 7 days in
+/// milliseconds. Without this cap such entries would be stored forever
+/// if the recipient never becomes reachable.
+const V2_MAX_RETENTION_MS: u64 = 7 * 24 * 60 * 60 * 1000;
```
MathJud, line 100: This value should be configurable via `config.yaml`
There should be more values, see discussion file:
- max custody time for users
- max custody time for trusted users

Issue #904, opened 2026-06-30
release preparation: Android 11, OnePlus 6T crashes at startup


MathJud:
- Device: OnePlus 6T
- Android version: 11
- Flutter version: Flutter 3.41.9 • channel stable

App panics at startup

flutter crash log:

```
[   +2 ms] Syncing files to device ONEPLUS A6013...
[   +2 ms] Compiling dart to kernel with 0 updated files
[        ] Processing bundle.
[   +1 ms] <- recompile package:qaul_ui/main.dart 02b2b39f-60ba-4d50-bf85-c131ecbc760f
[        ] <- 02b2b39f-60ba-4d50-bf85-c131ecbc760f
[   +1 ms] Bundle processing done.
[ +178 ms] Updating files.
[        ] Pending asset builds completed. Writing dirty entries.
[        ] DevFS: Sync finished
[        ] Syncing files to device ONEPLUS A6013... (completed in 185ms)
[   +1 ms] Synced 0.0MB.
[   +1 ms] <- accept
[   +5 ms] Connected to _flutterView/0x7a92b952b0.
[   +3 ms] Flutter run key commands.
[   +1 ms] r Hot reload. 🔥🔥🔥
[        ] R Hot restart.
[        ] h List all available interactive commands.
[        ] d Detach (terminate "flutter run" but leave application running).
[        ] c Clear the screen
[        ] q Quit (terminate the application on the device).
[        ] A Dart VM Service on ONEPLUS A6013 is available at: http://127.0.0.1:38087/PHcSoYpvzik=/
[   +1 ms] The Flutter DevTools debugger and profiler on ONEPLUS A6013 is available at:
http://127.0.0.1:38087/PHcSoYpvzik=/devtools/?uri=ws://127.0.0.1:38087/PHcSoYpvzik=/ws
[ +348 ms] W/Choreographer(24887): Already have a pending vsync event.  There should only be one at a time.
[   +1 ms] I/Choreographer(24887): Skipped 64 frames!  The application may be doing too much work on its main thread.
[  +31 ms] I/System.out(24887): Initialize libqaul with storage path: /data/user/0/net.qaul.qaul_app/files
[  +29 ms] F/libc    (24887): Fatal signal 6 (SIGABRT), code -1 (SI_QUEUE) in tid 24887 (t.qaul.qaul_app), pid 24887 (t.qaul.qaul_app)
[ +400 ms] *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
[...]

PR #909, opened 2026-07-07
fix(android): set TMPDIR so libqaul boots on all devices

n-0-a:
## Description

sled temporary(true) databases open in the system temp dir, which Rust's std::env::temp_dir() resolves to /tmp when TMPDIR is unset. Android never sets TMPDIR and /tmp is not writable on many OEM builds (Realme, Moto, Samsung), so startup panicked with EACCES on those devices while working fine on stock Android (Pixel). Point TMPDIR at the app's writable files dir, before the worker thread spawns, which covers all temporary(true) opens (messaging, crypto, dtn, feed, database) at once.


MathJud:
fixes #904

PR #911, opened 2026-07-07
Improve chat bubble layout and text scaling

GustavoDcs:
## Description
I tested the current main branch and could not reproduce the reported issue.

On my Android test device, chat message text and timestamps looked correct, and Android font/accessibility scaling was also being applied as expected in the app. I did not find evidence that the reported behavior is currently broken on main.

That said, I added a small safeguard in the chat bubble component: message text, timestamps, sender labels, and the bubble layout measurements now use the same capped text scaler. This should prevent future regressions where one text element scales and another does not, while still keeping the layout within the designer’s intended bounds.

## Evidence

Android Emulator (main):
Screenshot 2026-07-07 at 14 34 09
Screenshot_20260707_142025

Android Emulator (fix branch):
Screenshot_20260707_143903
Screenshot_20260707_143809

Android Device (Redmi 9i)

Scalled Text
Screenshot_2026-07-07-15-47-40-732_net qaul qaul_app

### DIFF
--- a/qaul_ui/packages/qaul_components/lib/design_components/chat/chat_footer.dart
+++ b/qaul_ui/packages/qaul_components/lib/design_components/chat/chat_footer.dart
@@ -485,7 +485,7 @@ class _ComposerPill extends StatelessWidget {
     final text = controller.text;
     if (text.contains('\n') || text.contains('\r')) return true;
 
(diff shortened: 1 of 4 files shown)


brenodt (approved):
LGTM

PR #920, opened 2026-07-09
Add group invite to notificatino count

GustavoDcs:
## Description
Fixed the chat tab notification badge so newly received group invites are counted as new chat activity.

The chat notification controller now listens for new group invites in addition to unread chat room updates, increments the main tab bar chat counter for new invites, and keeps the existing behavior of not counting while the user is already on the chat tab. Added regression tests for both cases.

## Evidence
https://github.com/user-attachments/assets/8ced8d65-17b0-4e3a-a4c7-f14d7dd118f5

### DIFF
--- a/qaul_ui/lib/providers/notification_controller/chat_notification_controller_provider.dart
+++ b/qaul_ui/lib/providers/notification_controller/chat_notification_controller_provider.dart
@@ -23,6 +23,7 @@ class ChatNotificationController extends NotificationController<List<ChatRoom>>
   @override
   Future<void> initialize() async {
     await super.initialize();
+    ref.listen(groupInvitesProvider, _executeGroupInvites);
     if (preferences.containsKey(cacheKey)) {
       _chats.addAll(
         preferences.getStringList(cacheKey)!.map((e) {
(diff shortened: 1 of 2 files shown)


qaul_ui/lib/providers/notification_controller/chat_notification_controller_provider.dart
```diff
@@ -36,6 +37,24 @@ class ChatNotificationController extends NotificationController<List<ChatRoom>>
     );
   }
 
+  void _executeGroupInvites(
```
brenodt, line 40: Nit: I'd call this `_processNewGroupInvite`

qaul_ui/lib/providers/notification_controller/chat_notification_controller_provider.dart
```diff
@@ -36,6 +37,24 @@ class ChatNotificationController extends NotificationController<List<ChatRoom>>
...
+    List<GroupInvite>? previous,
+    List<GroupInvite> current,
+  ) {
+    final previousInvites = previous ?? const <GroupInvite>[];
```
brenodt, line 44: We could make this method a bit more readable by making use of sets and checking the difference.

We also move the return early so it avoids allocating when unnecessary:

```dart
if (currentVisibleHomeTab == TabType.chat ||
    !UserPrefsHelper.instance.chatNotificationsEnabled) {
  return;
}
final newInvites = current.toSet().difference((previous ?? []).toSet());
if (newInvites.isEmpty) return;
```

brenodt (commented):
Not blocking on the nits, take it or leave it. If you agree, please open a new PR @GustavoDcs

PR #926, opened 2026-07-21
Rename and optimize group invite processing

GustavoDcs:
## Description
Improves group invite notification handling by detecting newly received invites with set difference and skipping unnecessary processing when chat notifications are disabled or the chat tab is already visible.

Related to the comments on #920


brenodt (approved):
LGTM

PR #929, opened 2026-07-24
refactor(qauld-tui): fetch in a background task, off the render loop

dastansam:
Stacked on #928 — set to merge into `fix/qauld-tui-dtn-fetch`; GitHub will retarget this to `main` automatically once #928 merges.

Polling used to run inline in the render loop's select tick — 8 RPCs awaited sequentially, freezing input/redraws while they ran and up to ~40s/cycle if the daemon was unreachable. Move polling to a background task: `refresh_once` fans the fetches out concurrently (one round-trip's latency, not eight) and posts a Send-able `Snapshot` to the loop, which applies it synchronously. No network `.await` anywhere in the loop. `r` and post-feed-send just fire a trigger.

Crypto events now dedup through one path (the poll floor lives with the task), so poll/push overlaps and the daemon's inclusive `since_ms` re-sends are dropped. Three new dedup unit tests.


MathJud:
I merged #935, which is this PR without the last commit, which contains the demo script.
The demo script has issues which should be resolved.
I suggest to move the demo script with the explanation to `utilities/demo/`

MathJud:
superseeded by #945

PR #937, opened 2026-08-06
Fix changing account log in chat issues

GustavoDcs:
## Description
Fixes stale session state after logging out and signing in with a different account.

The issue was that some Riverpod providers/stores kept data from the previous session after logout. When a new account authenticated, the UI could still show old chats/users, open invalid conversations, and try to send messages to groups the new user could not access, causing Group not found errors.

Changes included:
- Centralized session-state reset on logout, login/restore, and account creation.
- Clearing chat, user, invite, open chat, feed, and home-related providers when switching sessions.
- Rehydrating defaultUserProvider with the active user after login/account creation.
- Creating direct chats with a conversationId computed from the active local user instead of stale cached user data.
- Treating newly opened direct chats as drafts to avoid premature unread/refresh calls before the backend group exists.
- Adding initial loading states for chats/users to avoid empty-state flashes while the new session loads.
- Adding focused tests for session reset and direct chat id generation.
## Evidence
Before:

https://github.com/user-attachments/assets/1875fcf9-8936-43da-b507-207646051d22

After:

https://github.com/user-attachments/assets/19dc4d8f-0b39-41b3-a692-f163e673db31

### DIFF
--- a/qaul_ui/lib/coordinators/account_management_coordinator.dart
+++ b/qaul_ui/lib/coordinators/account_management_coordinator.dart
@@ -1,5 +1,3 @@
-import 'dart:async';
-
 import 'package:file_picker/file_picker.dart';
 import 'package:flutter/material.dart';
 import 'package:hooks_riverpod/hooks_riverpod.dart';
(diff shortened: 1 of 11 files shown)


brenodt:
Replaced by #940

PR #952, opened 2026-08-19
Forward message flow

GustavoDcs:
## Description
Adds the in-app forward message flow for text messages.

Users can long-press a text message, see the full design-system context menu with only Forward enabled, choose a single group or user recipient, and return to that chat with the forwarded text pre-filled in the ChatFooter composer. The user still has to manually send the message.

This also wires the design-system ChatFooter into the real chat screen and keeps existing text, attachment, camera, and audio entry points connected where supported.

## Evidence

https://github.com/user-attachments/assets/762e4385-9373-4dd8-9659-03453ecf2659


brenodt:
@GustavoDcs , could you rebase this one whenever you have some time? Thanks!

PR #953, opened 2026-08-21
feat(dtn-v2): signed immutable route, recipient grants + PoW admission, signed responses

dastansam:
Redesigns the DTN v2 custody protocol around a **signed, immutable route** and a **capability-based admission** model, replacing the flat, mutable, unauthenticated `DtnRoutedV2`. Implements the direction settled in the structures/policy feedback plus the grant-based admission fix.

Since the old flat `DtnRoutedV2` is opt-in and not in real use, the wire format is replaced outright — no migration.

## Wire format (`messaging.proto`)
- `DtnV2Container{ dtn_route, dtn_route_sig, envelope, custody_grant?, pow? }` — the route is a signed blob, immutable in transit.
- `DtnRoute{ original_signature, route_hop[], sender_public_key, expires_at? }` with `RouteHop{ route_entry[] }` / `RouteEntry{ id }` — position gives order, no cursor, no handoff counter.
- `CustodyGrant` (recipient-signed capability) and `PowStamp`.
- `DtnResponseV2` — **signed**, four kinds: `DELIVERY`, `CUSTODY_RELEASE`, `RECEIPT`, `DROP_REPORT`. Replaces the unauthenticated `DtnResponse` for v2.

## Custody pipeline (`services/dtn`)
- **Signed route:** signed on originate; custodians verify `dtn_route_sig` against the sender key and never rewrite the route.
- **Stateless traversal:** a custodian locates itself in the route and forwards to the furthest reachable hop past its own position (skipping dead hops); delivers directly when the recipient is reachable.
- **Admission is capability-first:**
  - a recipient-signed **grant** (`grantee`/`recipient`/`quota` checked), else
  - a locally **trusted** contact, else
  - a **proof-of-work** stamp bound to `(message, custodian, day)` for the grant-less untrusted pool, with an **aggregate pool cap** that actually bounds a Sybil flood (per-sender quota alone can't, since identities are free).
- **Signed responses:** only a validly-signed `DtnResponseV2` can free custody storage — fixes the forgeable v1 ACK where any peer could free or falsely confirm a stored message. Includes the reverse-facing DELIVERY ack, custody release, per-accept receipts, and drop-reports.
- **Tier-aware retention:** untrusted entries shed first / soonest.
- **Grant issue/import RPCs** (`dtn_rpc.proto`).

## Tests
- 29 new unit tests: route sign/verify + tamper, grant verify matrix, PoW (solve/verify, min-difficulty, custodian/day binding), signed-response verify + **forged-response-rejected**, stateless traversal (furthest reachable hop), end-to-end admission tiers, delivery-ack frees custody.
- Integration tests rewritten for the new wire format.
- Full workspace builds clean; 136 libqaul unit tests + 8 integration tests pass.

## Follow-ups (not blocking)
- `TODO(dtn-v2)`: route the DELIVERY ack back over the reverse of the forward route so it survives an **offline sender** (currently sent direct to the sender id).
- Only the Rust protobuf bindings are regenerated here; other-language bindings (cpp/java/kotlin/swift/dart) can be regenerated if/when the UI needs the grant RPCs.

### DIFF
--- a/protobuf/generated/rust/qaul.net.messaging.rs
+++ b/protobuf/generated/rust/qaul.net.messaging.rs
@@ -30,7 +30,7 @@ pub struct Envelope {
 /// envelop payload
 #[derive(Clone, PartialEq, ::prost::Message)]
 pub struct EnvelopPayload {
-    #[prost(oneof = "envelop_payload::Payload", tags = "1, 2, 3")]
+    #[prost(oneof = "envelop_payload::Payload", tags = "1, 2, 3, 4")]
     pub payload: ::core::option::Option<envelop_payload::Payload>,
 }
 /// Nested message and enum types in `EnvelopPayload`.
@@ -46,6 +46,10 @@ pub mod envelop_payload {
         /// directed custody routed DTN message (V2, signed immutable route)
         #[prost(message, tag = "3")]
         DtnV2(super::DtnV2Container),
+        /// a signed DTN v2 response carried as a custody payload, so a DELIVERY
+        /// ack can be routed back to an offline sender over the reverse route
+        #[prost(message, tag = "4")]
+        DtnResponseV2(super::DtnResponseV2),
     }
 }
 /// encrypted message data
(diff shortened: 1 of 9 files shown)


MathJud:
Continues #988 

PR #964, opened 2026-08-28
Adjust sender id wierd behaviour and edge case

GustavoDcs:
## Description
This PR fixes inconsistent user identity handling across chat and public feed navigation.

Direct chats now compute the conversation id from the currently active local user and the selected peer, instead of trusting a possibly stale conversationId stored on the peer model. This prevents newly opened direct chats from targeting an invalid or previous-session conversation id after account switches or newly created users.
It also normalizes qaul full user ids and q8 chat sender ids before comparing identities. Without this, messages could initially render as sent by the local user, then appear as sent by another user after the conversation was reloaded from backend state.

Public feed author navigation was adjusted so tapping your own author name/avatar opens the account profile tab, while tapping other authors still opens the user details screen.

Regression coverage was added for q8/full-id identity matching, direct chat id calculation from the active account, and feed author navigation for both self and other users.

### DIFF
--- a/qaul_ui/lib/decorators/search_user_decorator.dart
+++ b/qaul_ui/lib/decorators/search_user_decorator.dart
@@ -1,4 +1,3 @@
-import 'package:collection/collection.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_hooks/flutter_hooks.dart';
 import 'package:hooks_riverpod/hooks_riverpod.dart';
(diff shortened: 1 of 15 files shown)


brenodt:
I'd rather we tackle this on libqaul directly, otherwise other clients (like qauldctl) would still have the same sort of bug. So I'm closing this one in favor of #976 !

PR #969, opened 2026-08-29
fix(router): propagate profile updates to already-known peers

dastansam:
## The bug

A signed profile update (name / avatar / bio → `version` bump, re-signed) reaches **new** peers but never reaches peers who **already know** the user. Existing contacts see a stale profile forever — rename yourself and everyone who already has you keeps the old name.

## Root cause

Confirmed by tracing the full path:

- The periodic routing sync only re-requests **unknown** users — `Users::get_missed_ids` filters purely on `!contains_key` (`router/users.rs`), and the routing announcement (`RoutingInfoEntry`) carries **no version** (`router/table.rs`), so a version bump on a known user is invisible to a receiver.
- Signed profiles are only ever sent in response to a `UserRequest`, which only fires for missed (unknown) ids — so a known user's bumped profile is never re-requested.
- The **accept** side is already correct: `add_signed_user_info_table` replaces on `version > existing || (== && updated_at >)`. So the defect is purely on the **send/select** side.

## The fix

Minimal and contained — **no changes to the routing hot path / pgid state machine**:

- `RoutingInfoEntry` gains a `version` field (proto3, backward compatible — absent decodes to `0`).
- Outgoing announcements are stamped with the profile version from the authoritative Users store (`fill_routing_versions`), looked up at send time.
- The receiver re-requests users that are unknown **or** whose advertised version is newer than the one held — `get_stale_ids` replaces `get_missed_ids`.

**Multi-hop works for free:** accepting a newer profile already updates the Users store, so a node's next announcement carries the new version and the pull cascades hop by hop. **Low risk:** `version` only gates *which profiles to re-request* — it never affects routing decisions, so worst case is a missed or extra profile pull, never a broken route.

## Backward compatibility

An un-upgraded peer sends `version = 0`; a known user at any local version is then never spuriously re-requested (same as today). Unknown users are still always requested. No flag-day.

## Tests

Adds `stale_selection_requests_outdated_known_user` (reproducing test): a known user whose advertised version increased **must** be re-requested; a same-version known user must not; an unknown user still is. `router::users` suite: 17 passed.

### DIFF
--- a/protobuf/generated/rust/qaul.net.router_net_info.rs
+++ b/protobuf/generated/rust/qaul.net.router_net_info.rs
@@ -62,6 +62,11 @@ pub struct RoutingInfoEntry {
     /// propagation id
     #[prost(uint32, tag = "5")]
     pub pgid: u32,
+    /// advertised signed-profile version, so a receiver can detect that an
+    /// already-known user's profile changed and re-request it. 0 = unknown /
+    /// legacy peer. Backward compatible: an absent field decodes to 0.
+    #[prost(uint32, tag = "6")]
+    pub version: u32,
 }
 /// User information table
 #[derive(Clone, PartialEq, Eq, Hash, ::prost::Message)]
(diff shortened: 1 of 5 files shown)


MathJud:
Continues #987 

PR #970, opened 2026-08-29
fix(storage): atomic config write — prevent node identity/account loss on interrupted write

dastansam:
## Bug (HIGH — silent permanent data loss)
An interrupted config write (crash / kill / disk-full) leaves `config.yaml` truncated. On next boot `load_or_create` falls back to `Configuration::default()` over it; `Node::new` sees `initialized == 0`, **generates a new node identity and drops every user account**, then saves — overwriting the truncated file. One interrupted write permanently destroys the node's identity and all local accounts.

## Fix
- `save_to_path`: write to `config.yaml.tmp` then **atomically rename** into place — `config.yaml` is always the complete old or new file, never partial. Errors are logged, not `expect()` panics.
- `load_or_create`: a *present-but-unparseable* config is renamed to `config.yaml.corrupt` (covering both the build/truncation path and the deserialize path) instead of being silently discarded, so identity/accounts stay recoverable. A genuinely missing file is still normal first-run.

## Tests
`corrupt_config_is_backed_up_not_silently_lost`, `save_is_atomic_and_round_trips`. Full libqaul suite: 146 passed.

### DIFF
--- a/rust/libqaul/src/storage/configuration.rs
+++ b/rust/libqaul/src/storage/configuration.rs
@@ -384,18 +384,46 @@ impl Configuration {
         let path = Path::new(storage_path);
         let config_path = path.join("config.yaml");
 
+        // A present-but-unparseable config must NOT be silently discarded: the
+        // default path regenerates the node identity and drops every user
+        // account (see `Node::new`, which regenerates when `initialized == 0`).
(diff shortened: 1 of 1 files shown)


MathJud:
Continues #986 

PR #972, opened 2026-08-29
fix(storage): double-checked lock in get_user_db — avoid concurrent-open panic/poison

dastansam:
## Bug (HIGH — lock-poison outage)
`get_user_db` misses the account under a read lock, then takes the write lock and opens the sled DB **without re-checking**. Two threads first-opening the same account both call `sled::open` on one path; sled's exclusive per-process flock rejects the second, whose `.expect()` panics **while holding the storage write lock** — poisoning the `RwLock` so every later storage access panics (node-wide storage outage). Reachable when two RPC/network threads first touch the same freshly-added account.

## Fix
Re-check the open-handles map under the write lock (double-checked locking) and return the already-open handle, so only one `sled::open` ever runs per account.

## Tests
`concurrent_first_open_does_not_poison_storage_lock` — 8 threads first-opening the same account concurrently don't panic/poison. Full suite: 146 passed.

### DIFF
--- a/rust/libqaul/src/storage/database.rs
+++ b/rust/libqaul/src/storage/database.rs
@@ -136,6 +136,17 @@ impl DataBase {
             // get data base structure
             let mut database = state.database.inner.write().unwrap();
 
+            // Re-check under the write lock (double-checked locking): another
+            // thread may have opened this account's DB between our read-lock miss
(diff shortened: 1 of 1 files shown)


MathJud:
Continues #985

PR #974, opened 2026-08-31
fix: exit code 1 on data upgrade failure

lucasew:
Signed-off-by: lucasew <lucas59356@gmail.com>

## Description
The problem:
- we run qauld on systemd
- systemd restart failing services
- a failure was being reported as success (exit code 0)
- this fixes that


MathJud:
Thanks a lot for this fix! @lucasew

PR #976, opened 2026-09-12
fix(libqaul): resolve the signed-in account instead of config index 0

brenodt:


MathJud:
Continues #984 

